PR #6, opened 2022-07-23
Remove unnecessary workflow

rgreinho:
Removes unnecessary workflow.


rgreinho:
Force merge since it only removes a workflow. The true CI fix is in #7.

PR #162, opened 2023-04-20
Update BNA Development Documentation

lalver1:
# Pull-Request

## Types of changes

- Documentation

## Description

Adds documentation on how to develop the BNA using brokenspoke-analyzer to prepare and run the BNA.


docs/source/pfb-network-connectivity.rst
```diff
@@ -0,0 +1,101 @@
...
+	poetry shell
+
+Then, to install dependencies needed for project, run::
+
+	poetry install
+
+Finally, to exit the virtual environment created by poetry, run::
+
+	exit
+
```
rgreinho, line 96: These steps are unnecessary. 

`poetry install` is the command that creates the virualenv. So you just need to run this one.

`poetry shell` is the equivalent of runing `source .venv/bin/activate` then running your commands.

rgreinho (changes requested):
It looks really good @lalver1 !

Aside of a minor setup adjustment, I have a question regarding the organization: based on how you see this document evolving, do you think it should stay in its own page, or should we add it to the `CONTRIBUTING` page?

docs/source/pfb-network-connectivity.rst
lalver1: Thanks @rgreinho ! I don't think the document will grow substantially from what we have now, so I think I could just add it to the `CONTRIBUTING` page. I may reorganize `CONTRIBUTING` a bit to make the distinction clear between brokenspoke-analyzer and the BNA, but I don't expect any major changes. And I agree with the `poetry` setup steps, I'll update that section 👍.

lalver1: @rgreinho , is it ok if I change `CONTRIBUTING` to an `.rst` file so that the `graphviz` figure can be included?

rgreinho: I think so. You can try and if it breaks something you'll just have to submit another patch 😆

docs/source/conf.py
```diff
@@ -56,6 +55,9 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = []
 
+# Generate labels for heading anchors for h1, h2, and h3 level headings 
+# (corresponding to #, ##, and ### in markdown).
+myst_heading_anchors = 3
```
rgreinho, line 60: This was the tricky settings! 😮

rgreinho (approved):
Awesome work Luis! 👍

Issue #188, opened 2023-06-10
Invalid arguments for bna prepare


rgreinho:
# Bug report

The documentation states that we can run the following command:

```bash
bna prepare usa flagstaff arizona
```

However, this command fails when executed.

## Current Behavior

The command ails with the following error message:

```py
ValueError: The dataset 'usa, arizona' is not available.
```

## Expected Behavior

This command should run identically to:

```bash
bna prepare arizona flagstaff 
```

## Possible Solution

The command fails as it cannot find the dataset for the pair "country, state". This is due to the fact that Geofabrik does not have a dataset for this. It must either be USA, or Arizona.

Therefore a solution could be to ignore the country if the state is specified. In this case, it would lookup for only for the arizona dataset.

PR #190, opened 2023-06-10
Fix invalid dataset error in prepare

rgreinho:
Fixes the dataset error when specifying a country **AND** a state in the
prepare command.

Fixes PeopleForBikes/brokenspoke-analyzer#188


lalver1 (approved):
👍

PR #191, opened 2023-06-10
Make output_dir an option for the prepare command

rgreinho:
Converts `output_dir` from an optional argument to an option. This
allows the user to specify the output directory without having to
specify a state.

Fixes PeopleForBikes/brokenspoke-analyzer#190

Signed-off-by: Rémy Greinhofer <remy.greinhofer@gmail.com>

PR #192, opened 2023-06-10
Update the prepare command to gather US files

rgreinho:
Updates the logic of the `prepare` command to gather the files required
for the US in case of an american city.

These files were previously ignored since the BNA has some specific
logic to fetch the for American cities. However, in order to make the
BNA more generic, it makes sense to enhance this command to make it
exhaustive.


rgreinho:
Must be merged after #190 and #191.

lalver1 (approved):
👍

PR #205, opened 2023-06-17
Fix Geofabrik downloader for Spain

rgreinho:
Fixes the Geofabrik downloader for Spain.

There are 3 problems with the current implementation:

- a trailing slash ('/') is missing for the spanish URL
- Spain is not added as a European subregion
- the Spanish regions use mix-case but should be all lower case


lalver1 (approved):
👍

PR #206, opened 2023-06-17
Adjust synthetic population shapefile name

rgreinho:
Renames the population shapefile generated for insternational. The goal
is to be consistent between the US cities and the international cities.
Therefore the census shapefiles and the synthetic population shapefile
must have the same name.


lalver1 (approved):
👍

PR #214, opened 2023-06-24
Sanitize Docker variables

rgreinho:
Docker variables cannot contain spaces, therefore we must replace them
with underscores (`_`).

Drive-by:
- Removes trailing spaces in URLs.


lalver1 (approved):
👍

PR #223, opened 2023-07-01
Add California dataset

rgreinho:
The OSM dataset for the state of California was only available through
"Northern California" or "Southern California", but as these are not
official states the requests for any city in California were failing.


lalver1 (approved):
👍

PR #225, opened 2023-07-01
Ensure regions use only ASCII characters

rgreinho:
The region files are downloaded from Geofabrik or BBike, and their names
are written using only ascii characters. Submitting a request to
download a region with a unicode character will result in a failure.
This patch ensures that the region names are submitted only with ASCII
characters.


lalver1 (approved):
👍

Issue #258, opened 2023-07-31
Remove dependency to GDAL


rgreinho:
# Feature request

## Current Behavior

We currently use GDAL to determine the UTM zone of a shapefile. However it is very complicated to install in an unattended fashion and the python bindings depend on C libraries which are different on every system, making the setup experience even worse.

## Expected Behavior

The setup should be simpler and the end user should not have to go through so many steps just to be able to determine the UTM zzone of a shape file.

## Possible Solution

Use the Geopandas `estimate_utm_crs` function instead (https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.estimate_utm_crs.html)

```py
>>> import geopandas as gpd
>>> gdf = gpd.read_file("santa-rosa-new-mexico-usa.shp")
>>> gdf.crs
<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: World.
- bounds: (-180.0, -90.0, 180.0, 90.0)
Datum: World Geodetic System 1984 ensemble
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich

>>> gdf.estimate_utm_crs()
<Projected CRS: EPSG:32613>
Name: WGS 84 / UTM zone 13N
Axis Info [cartesian]:
- E[east]: Easting (metre)
- N[north]: Northing (metre)
Area of Use:
- name: Between 108°W and 102°W, northern hemisphere between equator and 84°N, onshore and offshore. Canada - Northwest Territories (NWT); Nunavut; Saskatchewan. Mexico. United States (USA).
- bounds: (-108.0, 0.0, -102.0, 84.0)
Coordinate Operation:
- name: UTM zone 13N
- method: Transverse Mercator
Datum: World Geodetic System 1984 ensemble
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich

>>> utm = gdf.estimate_utm_crs()
>>> utm.utm_zone
'13N'
>>> utm.to_string()
'EPSG:32613'
>>> utm.to_string()[5:]
'32613'
```

Issue #248, opened 2023-07-29
Import the new functions from the modular BNA.


rgreinho:
# Feature request

## Expected Behavior

Reorganize the API based on what was added during the modular BNA experiment.

PR #259, opened 2023-07-31
Remove GDAL dependency

rgreinho:
Removes the GDAL dependencies by using Geopandas functions instead. This
greatly simplify the installation process.

Fixes PeopleForBikes/brokenspoke-analyzer#258.


rgreinho:
Must be merged after #248.

lalver1:
I had no problems installing using Poetry @rgreinho . The prepare command also ran without a problem so I think the GDAL replacement, `get_srid()`, works well. The reorganization also makes sense, I don't see any issues 👍.

PR #271, opened 2023-08-11
Delete dwnld retry

lalver1:
# Pull-Request

## Types of changes

## Description

Adds retrying capability and deletes downloaded files if an exception occurs during download.
Tested by calling from modular BNA. Prepare steps passed but I got an error in the export step of the modular BNA. I believe the error is not related to changes in this PR.


rgreinho:
Superseeded by #272.

Issue #222, opened 2023-07-01
Retry on failure from a prepare step


rgreinho:
# Feature request

## Current Behavior

If a prepare step fails it does not retry automatically.

## Expected Behavior

If a  step from `prepare` fails, the brokenspoke-analyzer should retry automatically as this is most likely a network error. 

This should work for both the `prepare` and the `run` command.

## Possible Solution

Use [tenacity](https://tenacity.readthedocs.io/en/latest/#) to configure some retry logic around the download functions.

The number of attempts should be configurable at runtime (https://tenacity.readthedocs.io/en/latest/#changing-arguments-at-run-time), from the command line. Possible the delay as well. For instance:

```bash
bna --retry 2 --retry-delay 10 prepare arizona flagstaff
```

Issue #221, opened 2023-07-01
Partial downloads should be deleted


rgreinho:
# Bug report

## Current Behavior

When running `bna prepare` or `bna run` and the download of one of the input files fails, the file stays on disk. If the command is run a second time, the file will be considered as downloaded even though it is incomplete, causing the step to be skipped.

## Expected Behavior

If the file was not download entirely, if should be deleted from the disk.

## Steps to Reproduce

Run `bna prepare arizona flagstaff`, and cancel the process with CTRL+C in the middle of the download (or turn off the internet or wifi).

PR #272, opened 2023-08-12
Remove partial downloads and add retries

lalver1:
If an error occurs when downloading (lodes data, census waterblocks, census blocks, state speed limit, city speed limit, and region file), the partial downloads will be deleted. Retrying download capability is also added to the cli.

Fixes PeopleForBikes/brokenspoke-analyzer#221
Fixes PeopleForBikes/brokenspoke-analyzer#222


rgreinho (approved):
👍

Issue #247, opened 2023-07-29
Reorganize the cli


rgreinho:
# Feature request

## Expected Behavior

Create a dedicated package for the CLI, and a dedicated module for each sub-command:

```bash
cli
├──__init__.py
├── root.py
├── prepare.py
├── import.py
├── compute.py
├── export.py
└── run.py 
```

Refer to the [Typer documentation](https://typer.tiangolo.com/tutorial/subcommands/add-typer/) to see how to proceed to organize the sub-commands.

The commands are the following ones:

- Prepare: prepare the necessary files to run the analysis
- Import: import the files into PostgreSQL/PostGIS
- Compute: perform the analysis
- Export: export the results. 
- Run: a combination of all the previous steps
  - With or without managing docker-compose
  - Keep the simplicity: bna run <country> <city> [<state> [<city_fips>]]
    - US cities must specify the state and the city FIPS code.
    - For non-US cities:
      - The FIPS code is always ignored and set to 0000000.
      - The state is optional, and useful only if there are several cities with the same name in different parts of the country or to retrieve a specific (sub)region (it is used only to retrieve the boundary file and the region file).
    - We want to keep the ability to run the original BNA as well

Other requirements:

- The CLI should be just a thin wrapper passing parameters to core functions and orchestrating them.
- Use smart default values
  - Better let it fail than producing incorrect results.
- The CLI must display user friendly error messages by default. If opted-in it should show the stack trace.

PR #274, opened 2023-08-14
Reorganize CLI

lalver1:
Reorganizes the CLI with a dedicated module for the prepare and compute commands.

- Groups parameters that are common to the commands into `brokenspoke_analyzer/cli/__init__.py` to avoid editing different files when a parameter requires a change.
- Updates `brokenspoke_analyzer/main.py` to point to `brokenspoke_analyzer/cli/root.py` 
- Updates the Poetry scripts.

Working on reorganizing remaining commands.

Fixes https://github.com/PeopleForBikes/brokenspoke-analyzer/issues/247


lalver1:
@rgreinho I had to remove the `asyncio.run()` call in `importer.py` and `export.py` because Mypy kept giving a `[func-returns-value]` error when the underscored functions were wrapped with `asyncio.run()`. Now that I think about it, it's probably my mistake since I did not include `async` in the coroutine definition, I'll test this. But in any case, the import and export commands will probably not run asynchronous tasks, so the current commands are good enough to move forward I guess.

rgreinho (approved):
Fantastic @lalver1 ! That puts us in a great spot for the next steps. Great job 👍

PR #285, opened 2023-08-20
Attempt to fix poetry/numpy

rgreinho:
For some reason Poetry cannot install Numpy 1.25.2, so we're trying with the previous version.


lalver1 (approved):
:+1:

PR #301, opened 2023-09-01
Package as Docker container

rgreinho:
Packages the application as a Docker container.

This patch also includes the SQL scripts inside the python package
itself.

The runtime depepdencies must be exported to the `requirements.txt`
format in order to be used to build their respective wheels during the
Docker build process.


lalver1 (approved):
Worked great! :+1:

Issue #313, opened 2023-09-06
Adjust log output according to the log level


rgreinho:
# Feature request

## Current Behavior

The log output displays a lot of information that is not always necessary cased on the verbosity level the user wants.

## Expected Behavior

The log output should adjust dynamically based on the verbosity level.

## Possible Solution

- -v: INFO level / basic logging information / `message`
- -vv: DEBUG level / more detailed information / <level>{time:YYYY-MM-DDTHH:mm:ssZZ} {level:.3} {module}:{line} {message}</level>
- -vvv: TRACE level / trace information / <level>{time:YYYY-MM-DDTHH:mm:ssZZ} {level:.3} {name}:{line} {message}</level>

rgreinho:
@lalver1 solved this issue in #361.

rgreinho:
Sorry, I have been told in my earbud that #301 only enables the ability to interact with the logging level, not the logging format. Reopening the issue.

rgreinho:
Superseeded by #369.

PR #319, opened 2023-09-09
Add log for download retries

lalver1:
Adds the `before_log` feature to the download retrier to log retry
attempts.

Fixes PeopleForBikes/brokenspoke-analyzer/issues/290


brokenspoke_analyzer/cli/prepare.py
```diff
@@ -96,8 +99,16 @@ async def prepare_(
     # Prepare the Rich output.
     console = Console()
 
+    # Prepare logging for Retrying.
+    logging_logger = logging.getLogger(__name__)
+    logging.basicConfig(stream=sys.stderr, level=logging.WARNING)
```
rgreinho, line 104: This is correct, but since we use `loguru` as our logger throughout the app, we should use it here as well to be consistent.

rgreinho (changes requested):

brokenspoke_analyzer/cli/prepare.py
lalver1: Thanks and that makes sense @rgreinho. I originally was setting up the retryer like this:
`retryer = Retrying(
        stop=stop_after_attempt(retries),
        reraise=True,
        before=before_log(logger, 30),
    )`
but mypy throws an error:
`brokenspoke_analyzer/cli/prepare.py:104: error: Argument 1 to "before_log" has incompatible type "loguru.Logger"; expected "logging.Logger"  [arg-type]`
So I think I need to add something to tell mypy to ignore this line? Especially since the `prepare` command runs without a problem, it looks like the error is a mypy quirk?
Also, I've been trying to find the `loguru` equivalent to `logging.WARNING` (I know for loguru `30` is the severity value for `WARNING` per the docs), but I can't find an attribute anywhere in either `loguru` or `loguru.logger` that holds this number. So I'm currently setup the retrier like this:
`retryer = Retrying(
        stop=stop_after_attempt(retries),
        reraise=True,
        before=before_log(logger, 30),
    )` 
but there has to be a better way than just setting `30` in the function call, are you familiar where `logger` stores this `int`?

rgreinho: Ah yes, for mypy maybe we'll have to use `# type: ignore`.

For the log level in loguru, we control it via the CLI: https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/brokenspoke_analyzer/cli/root.py#L55.

lalver1: For mypy I tried 
`retryer = Retrying(
        stop=stop_after_attempt(retries),
        reraise=True,
        before=before_log(logger, logger.level("WARNING").no),
    )  # type: ignore`
but I still got the mypy error when linting.
I also modified `pyproject.toml` and added 
`[[tool.mypy.overrides]]`
`module = ["tenacity"]`
`ignore_errors = "true"`
as well as 
`[[tool.mypy.overrides]]`
`module = ["tenacity"]`
`disable_error_code = "arg-type"` 
but I still got the same error. So I finally had to add `# mypy: ignore-errors` to the beginning of `prepare.py` but this doesn't feel right :unamused:  

The log level is fixed though, it turns out that `logger.level("WARNING").no` holds the integer value of the log level. I also noticed that the callback was not being called in https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/c685c84a585af6d77cd4c9325cf6920a6e40866c/brokenspoke_analyzer/cli/root.py#L59 so I changed the line to `app = typer.Typer(callback=callback)`.

rgreinho: Excellent investigation work Watson 🕵️ 😁 

So a little bit of clarification:

- When using the `before_log` function, the second argument is the level with which you want to call the tenacity logging. Tenacity will write "Starting call to {fn_name}, this is the X time calling it", using the level you gave. So there is no need to give the same level as the logger. I would recommend using `logging.DEBUG`. Which means the message will only be displayed when using `bna -vv` since `-vv` means DEBUG level.
- The `# type: ignore` pragma, must go either at the end of the line, if the function is written on a single line, or next to the argument to ignore in case of a multiline.
- The configurations in the `pyproject.toml` are not necessary.
- The `# mypy: ignore-errors` at the top of the file neither.

So here is a sample:

```py
# Create retrier instance to use for all downloads
    retryer = Retrying(
        stop=stop_after_attempt(retries),
        reraise=True,
        before=before_log(logger, logging.DEBUG),  # type: ignore
    )
```

And for the callback, you are absolutely, it is currently not being used. I planned a fixed for v2.2 (https://github.com/PeopleForBikes/brokenspoke-analyzer/issues/313), but it you feel like it, you can fix it before. You already solved half of the problem.

lalver1: :grinning: thanks @rgreinho ! Your explanation about pragmas will help me a lot going forward! And the `before_log` explanation makes sense too :+1: 
I removed the callback again for consistency but I can definitely look at #313 after this PR is closed.

rgreinho (approved):

Issue #336, opened 2023-09-16
Having "state" and "region" parameters is confusing


rgreinho:
# Bug report

## Current Behavior

In the CLI we are using `state` most of the time and some times `region` as a parameter.

```bash
 Usage: bna run-with compose [OPTIONS] COUNTRY CITY [STATE] [FIPS_CODE]

 Usage: bna export local-calver [OPTIONS] COUNTRY CITY [REGION] [EXPORT_DIR]
```

## Expected Behavior

Since state mostly (only?) applies to the US, we should use the term `region` instead. This way it would work for cities worldwide.

PR #347, opened 2023-09-17
Use "region" keyword instead of "state"

rgreinho:
The usage of the word "state" applies mostly to the US. Since the
brokenspoke-analyzer is a worldwide application, we dediced to rename
the flags/fields to region to make it more applicable to all countries
of the world.

Fixes PeopleForBikes/brokenspoke-analyzer#336


brokenspoke_analyzer/cli/run_with.py
```diff
@@ -81,7 +81,7 @@ def compose(
 def original_bna(
     city_shp: pathlib.Path,
     pfb_osm_file: pathlib.Path,
-    state: typing.Optional[str] = None,
+    region: typing.Optional[str] = None,
```
lalver1, line 84: This does not affect functionality, but maybe change type to `common.Region` (`typing.Optional[str]`) to ensure consistency with the other `run-with` sub-commands.

lalver1, line 84:
```suggestion
    region: common.Region = None,
```

lalver1 (changes requested):
Looks good besides one minor comment.

lalver1 (approved):

Issue #331, opened 2023-09-16
Do not mix data and results


rgreinho:
# Feature request

## Current Behavior

Currently we are saving the data in `<output_dir>/data/<city-slug>` and and the results in a subdirectory of the data, like `<output_dir>/data/<city-slug>/results`.

## Expected Behavior

It would be better to separate them, like saving the data in `<output_dir>/data/<city-slug>` and the results in `<output_dir>/results/<city-calver>`.

## Possible Solution

The directory structure could look like this:

```
<output_dir>
├── data
│   └── santa-rosa-new-mexico-usa
└── results
    └── usa
        └── new mexico
            └── santa rosa
                └── 23.9 
```

PR #349, opened 2023-09-17
Separate data and results directories

rgreinho:
Instead of mixing the data and the results in the same directory, this
patch ensures that they are stored in 2 distincts directories.

Fixes PeopleForBikes/brokenspoke-analyzer#331


lalver1 (approved):
:+1:

Issue #334, opened 2023-09-16
Add a --version flag


rgreinho:
# Feature request

## Current Behavior

As a user I would like to be able to get the version number from the CLI.

## Expected Behavior

We should be able to run: 
```
$ bna --version
2.2.0
```

PR #361, opened 2023-09-24
Add "--version" Option flag

lalver1:
Added a "--version" option at the top level of the CLI. Modified the
original callback to support two Options: "--verbose" and "--version".

Fixes PeopleForBikes/brokenspoke-analyzer#334

### DIFF
--- a/brokenspoke_analyzer/cli/root.py
+++ b/brokenspoke_analyzer/cli/root.py
@@ -1,11 +1,16 @@
 import logging
 import pathlib
 import sys
-from importlib import resources
+from importlib import (
+    metadata,
+    resources,
+)
+from typing import Optional
 
 import typer
 from loguru import logger
 from rich.console import Console
+from typing_extensions import Annotated
 
 from brokenspoke_analyzer.cli import (
     common,
@@ -25,8 +30,7 @@ from brokenspoke_analyzer.core import (
 from brokenspoke_analyzer.core.database import dbcore
 
 
-def callback(verbose: int = typer.Option(0, "--verbose", "-v", count=True)) -> None:
-    """Define callback to configure global flags."""
+def _verbose_callback(value: int) -> None:
     # Configure the logger.
 
     # Remove any predefined logger.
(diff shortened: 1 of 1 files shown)


brokenspoke_analyzer/cli/root.py
```diff
@@ -55,8 +59,43 @@ def callback(verbose: int = typer.Option(0, "--verbose", "-v", count=True)) -> N
     logger.add(sys.stdout, format=log_format, level=log_level, colorize=True)
 
 
+def _version_callback(value: bool) -> None:
+    # Get the package's version
+
+    package_version = metadata.version("brokenspoke-analyzer")
+    if value:
+        print(f"brokenspoke-analyzer version: {package_version}")
```
rgreinho, line 67: Since we're using typer, we should be using `typer.echo` to display message:

```suggestion
        typer.echo(f"brokenspoke-analyzer version: {package_version}")
```

brokenspoke_analyzer/cli/root.py
```diff
@@ -55,8 +59,43 @@ def callback(verbose: int = typer.Option(0, "--verbose", "-v", count=True)) -> N
...
 # Create the CLI app.
 app = typer.Typer()
+
+
+@app.callback()
+def callback(
+    version: Annotated[
+        Optional[bool],
+        typer.Option(
+            "--version",
```
rgreinho, line 80: Can we also add the short version ('-v') and the count option? This way we could specify `bna -v` for lite logging or `bna -vvv` very verbose logging.

rgreinho (changes requested):
Great work Luis! 👍 2 little requests, but ready to merge it after that!

brokenspoke_analyzer/cli/root.py
lalver1: Ah yes! :+1:

brokenspoke_analyzer/cli/root.py
lalver1: No problem! I added the short version to the verbosity Option.

rgreinho:
Awesome @lalver1! Let's merge it!

rgreinho (approved):

lalver1:
Thanks @rgreinho !

Issue #300, opened 2023-08-30
Make PgAdmin optional in compose.


rgreinho:
# Feature request

## Current Behavior

Currently, docker compose runs PgAdmin, even when it is not required, for instance if one just wants to run an analysis and not inspecting the database.

## Expected Behavior

There should be 2 compose file, one to only run PostGIS, the only one to run both PostGIS and PgAdmin.

## Possible Solution

Refer to this blog article to get an overview of the ability to extend compose files: https://www.docker.com/blog/improve-docker-compose-modularity-with-include/

PR #364, opened 2023-09-27
Make pgAdmin optional

rgreinho:
Ensures the main Docker Compose file start only the required
PostgreSQL/PostGIS server, while keeping the ability to plug other
components like pgAdmin.

Fixes PeopleForBikes/brokenspoke-analyzer#300


lalver1 (approved):
:+1:

Issue #340, opened 2023-09-16
The CLI commands contain too much logic


rgreinho:
# Bug report

## Current Behavior

The CLI commands start to perform too many tasks before calling the API.

## Expected Behavior

The CLI commands should almost be only calling functions from the API.

rgreinho:
The CLI module should ideally not contain functions like `_local()`, `run_()`, etc. These function should belong to a module in the `core` package.

PR #366, opened 2023-09-27
Trim down the CLI

rgreinho:
Removes the excess of logic in the CLI commands.

Fixes PeopleForBikes/brokenspoke-analyzer#340


lalver1 (approved):
:+1:

Issue #273, opened 2023-08-13
Use SQLAlchemy ORM when applicable


rgreinho:
# Feature request

## Current Behavior

We use external SQL scripts to create some tables or perform some basic operations.

## Expected Behavior

Instead, we could leverage SQLAlchemy and its ORM.

Here are the candidates for the conversion:

- clip_osm.sql
- create_jobs_table.sql
- create_us_water_blocks_table.sql
- speed_tables.sql

There are also some tables created via `psql` in the code:

- the jobs table in the `ingestor.load_jobs()` function
- saving the speed limits into the `residential_speed_limit` table in the `ingestor.manage_speed_limits()` function
- validating the speed limit data  in the `ingestor.manage_speed_limits()` function

And the last piece would be to fix the `dbcore.import_csv_file_with_header()` function. An idea for this one would be to keep a version using `psql` for the import, but creating another version which would use the ORM to create the table if needed and insert the data into it.

rgreinho:
The first attempt was not very successful. We'll need to rethink this idea.

PR #367, opened 2023-09-28
Manage tables with SQLAlchemy

rgreinho:
Manages tables with SQLAlchemy ORM where it makes sense.

Fixes PeopleForBikes/brokenspoke-analyzer#273


rgreinho:
I am wondering if it is worth it just for 3 tables... 🤔

rgreinho:
It does not look like this is bringing a lot of value. We'll revise the idea later.

PR #369, opened 2023-09-30
Configure loguru to use RichHandler

rgreinho:
Configures the loguru sink to use the RichHandler instead of a custom
one.


lalver1 (approved):
:+1:

Issue #339, opened 2023-09-16
Add CI workflow to package the application


rgreinho:
# Feature request

## Expected Behavior

Add a CI workflow to create dist and wheel packages of the project. The packages should be attached to the GitHub release, as assets.

PR #382, opened 2023-10-15
Add sdist and wheel to the Release assets (#339)

lalver1:
Fixes PeopleForBikes/brokenspoke-analyzer#339

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,3 +14,21 @@ jobs:
     uses: PeopleForBikes/.github/.github/workflows/release-python.yml@main
   docker:
     uses: PeopleForBikes/.github/.github/workflows/docker-build-publish-ghcr.yml@main
+
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Gr1N/setup-poetry@v8
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: "poetry"
+      - name: Setup the project
+        run: poetry install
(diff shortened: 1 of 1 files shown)


lalver1:
I think this `release` job may conflict with the shared `release` job @rgreinho, especially with the lines
```
- name: Publish the release
  uses: softprops/action-gh-release@v1
  with:
    body: ${{ steps.changelog_reader.outputs.changes }}
```
but we can clean it up.

rgreinho:
Ah, I see. I did not think of it :/

So I went though the issues, and I found 2 that may be promising:
- https://github.com/softprops/action-gh-release/issues/203
- https://github.com/softprops/action-gh-release/issues/323

The first one shows that the `softprops/action-gh-release` action should upload assets to an existing release. The second one gives a workaround using the `gh` cli tool.

Could you try validating that the `softprops/action-gh-release` action will update an existing release?

lalver1:
Thanks for the tip @rgreinho ! I'll test `softprops/action-gh-release` for updating the assets of an existing release.

lalver1:
Looks like it works @rgreinho ! If the `softprops/action-gh-release` action is ran in a different workflow, any files included in the action are appended to an existing release. This is the simple repo I set up to test it: [gh-release-test](https://github.com/lalver1/gh-release-test). If you don't mind browsing over the 2 workflows (one mimics our shared `release`) just to make sure it reflects the actual workflows in `brokenspoke-analyzer` that would be good. Then I can update this PR. Thanks!

rgreinho:
Awesome! That looks great @lalver1! Merging 👍

rgreinho (approved):

PR #393, opened 2023-10-19
Add parameter to specify the S3 bucket name

rgreinho:
Instead of using a default hardcoded bucket name, this patch wires the
logic to use the bucket name provided via the CLI.


lalver1 (approved):
:+1:

PR #433, opened 2023-11-16
Review API docs

rgreinho:
Review and fix the docstrings and doctests.

The ruff linter for pydocstyle was enabled as well, and all the core
modules added to the API docs.


lalver1 (approved):
Looks good! :+1:

Issue #484, opened 2024-01-31
"'Until' is not recognized as an internal or external command"


jaredcowing:
# Bug report

Me again, I'm trying again to run a local analysis using the Poetry virtual environment and I get the error: "'Until' is not recognized as an internal or external command"
I am unsure if this is an issue with my setup that I need to sort out on my own or an issue with the code, but thought I would check in case you had a sense for the nature of the issue.

## Current Behavior

Once in the virtual environment I initiate the analysis with this command:
`bna run-with compose usa "pittsfield" "massachusetts" 25`

The traceback starts with this:

```
 C:\Users\jec15\Documents\GitHub\brokenspoke-analyzer\brokenspoke_analyzer\c |
| li\run_with.py:55 in compose                                                |
|                                                                             |
|    52     database_url = "postgresql://postgres:postgres@localhost:5432/pos |
|    53     try:                                                              |
|    54         subprocess.run(["docker", "compose", "up", "-d"], check=True) |
| >  55         subprocess.run(                                               |
|    56             f"until pg_isready -d {database_url}; do sleep 5; done",  |
|    57             shell=True,                                               |
|    58             check=True,     
```

## Context

This error was produced in a Windows environment using Git Bash.

rgreinho:
Ah yes, that's a problem on our end indeed! Thank you very much for reporting it. 

`until` is a bash keywork, mostly used on OSX or Linux, that's why it did not work for you. I will try to either move the logic to the `docker-compose.yml` file, maybe using a Healthcheck, or using a `docker run` command instead of the`until` one.

In the meantime, maybe using [gitbash](https://gitforwindows.org) or [cygwin](https://cygwin.com/install.html) would unblock you.

rgreinho:
@jaredcowing, we believe we fixed the error you reported in PR #486. If you still encounter problems, don't hesitate to re-open this issue.

## Pittsfield, MA

Here are some extra information regarding the analysis you are trying to run that we believe may help you.

First, the FIPS CODE of Pittsfield, MA is `2553960`.  You can find the value in this [document](https://www2.census.gov/geo/docs/reference/codes2020/place/st25_ma_place2020.txt) from the US Census 2020. We describe the process to find FIPS CODEs in our [documentation](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/docs/source/about.md#where-to-find-the-fips-codes) (soon to be online, when we release v2.0.0).

Therefore the command to run would be:

[...]

PR #486, opened 2024-02-03
Fix `run-with` command for Windows

rgreinho:
The `run-with` command cannot run on Windows as it uses some bash
instructions to ensure the Docker containers are up and healthy.

This patch moves the logic to Docker Compose itself to make the
analyzer platform independant.

Fixes PeopleForBikes/brokenspoke-analyzer#484


lalver1 (approved):
:+1:

PR #503, opened 2024-02-19
Update Dockerfile to use Python 3.12

rgreinho:
- Updates Dockerfile.
- Updates dependencies.


lalver1 (approved):
:+1: 
@rgreinho, I only noticed one thing that we can investigate later. In `compose.yml`, I had to change the `interval` parameter in the healthcheck from 10s to 15s, otherwise `configure_db` in `dbcore.py` would give a `psycopg.OperationalError`. But I guess this depends on how busy the host machine is when the bna runs, because last time 10s was enough.

PR #542, opened 2024-03-16
Update destinations OSM tags

lalver1:
Updated dentists, doctors, hospitals, pharmacies, retail, and schools to incorporate alternate or new OSM tags.


rgreinho (approved):
W00t! 👍

PR #571, opened 2024-04-11
Fix Calver Format

rgreinho:
The calver format was not using the zero-padded month number. This patch
ensures that the month number always has 2 digits.


lalver1 (approved):
Looks good! 👍

PR #603, opened 2024-05-13
Add missing param for the run command

rgreinho:
Adds the missing parameter allowing the run command to export the
results in a custom S3 directory in a custom bucket.


lalver1 (approved):
👍

PR #617, opened 2024-05-24
Add the ability to bundle the results

rgreinho:
Implements a function bundling the results into a zip file, and modify
the relevant export commands to support this feature.


lalver1 (approved):
Tested it and worked great!

PR #648, opened 2024-07-03
Change neighborhood_reachable_roads_low_stress to neighborhood_reachable_roads_high_stress in connected census blocks calculation

mitchellhenke:
# Pull-Request

Thank you for all of the work that goes into the ratings! I have been working on understanding how the scores work and running them locally and came across what I think is a bug.

## Types of changes

## Description

This changes the table used when calculating census block connection scores. The second clause seemed like it was intended to be using the `neighborhood_reachable_roads_high_stress` table with the TRUE `high_stress` value and `hs` alias for the table. I'm not 100% sure, but I figured this could impact calculated scores if that was the case, so I wanted to open this to confirm first.

I've run the unit tests locally and they passed. I have also tested it manually on some smaller cities I'm familiar with, and in general the scores seem to be lower with this change, but still near the results on https://cityratings.peopleforbikes.org/.

I'm very much interested in helping where I can with this tool, so if there's anything I can do, just let me know. Thanks again for making this!

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/connectivity/connected_census_blocks.sql
+++ b/brokenspoke_analyzer/scripts/sql/connectivity/connected_census_blocks.sql
@@ -34,7 +34,7 @@ SELECT
     TRUE, -- noqa: AL03
     (
         SELECT MIN(hs.total_cost)
-        FROM neighborhood_reachable_roads_low_stress AS hs
+        FROM neighborhood_reachable_roads_high_stress AS hs
         WHERE
             hs.base_road = ANY(source.road_ids)
             AND hs.target_road = ANY(target.road_ids)


lalver1:
Hi! We're so glad that you've been using the tool and thanks for the feedback! I agree, this does look like a bug, the second clause was likely meant to use `neighborhood_reachable_roads_high_stress table`. I'll look at this a little more closely and run a few cities before reviewing but my gut feeling is that we need to implement this change. Thanks again and talk to you soon!

lalver1 (approved):
This looks good! 👍 
If you can rebase and update the PR when you get a chance I'll go ahead and merge it.

mitchellhenke:
@lalver1 done!

lalver1:
👍

rgreinho (approved):
👍

PR #652, opened 2024-07-06
Fix Dockerfile

rgreinho:
Fixes the Dockerfile as GeoPandas 1.0.0 does not need Fiona as the I/O
engine, and as a result, we can get rid of the workaround that was
needed to install the brokenspoke analyzer.

Drive-by:
- Reorders the CI steps in the release workflow to ensure the Docker
  image builds before creating the release.


lalver1 (approved):
👍

PR #660, opened 2024-07-11
Reduce size of connected roads tables and indexes

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- [x] Code cleanup / Refactoring
- Documentation

## Description

In trying to run analysis on Milwaukee, WI, the `generated.neighborhood_reachable_roads_high_stress` table got pretty large and I couldn't finish it because I ran out of disk space trying to create the indexes.

The size of `generated.neighborhood_reachable_roads_high_stress` prior to creating the index was around 77GB:

|                Table Name                |                     Index Name                     | Total Size | Total Size of all Indexes | Table Size | Index Size | Estimated table row count |
|----------------------------------------- | -------------------------------------------------- | ---------- | ------------------------- | ---------- | ---------- | ------------------------- |
| neighborhood_reachable_ <br> roads_high_stress | neighborhood_reachable_ </br> roads_high_stress_pkey      | 73 GB      | 22 GB                     | 51 GB      | 22 GB      |                1055619328 |

<summary>
Index Query
<details>

```
SELECT i.relname "Table Name",indexrelname "Index Name", idx_tup_fetch, idx_tup_read,
pg_size_pretty(pg_total_relation_size(relid)) As "Total Size",
pg_size_pretty(pg_indexes_size(relid)) as "Total Size of all Indexes",
pg_size_pretty(pg_relation_size(relid)) as "Table Size",
pg_size_pretty(pg_relation_size(indexrelid)) "Index Size",
reltuples::bigint "Estimated table row count"
FROM pg_stat_all_indexes i JOIN pg_class c ON i.relid=c.oid ORDER BY pg_relation_size(relid) desc;
```

</details>
</summary>

A lot of the size is in the primary key column and index, which don't seem to be used. `neighborhood_connected_census_blocks` and `neighborhood_ways_net_link` are large-ish tables that could potentially also drop their primary key column.

It looked like the float column was only used as integers, so I changed it to an `int` as well since it uses less space and seems to ultimately be converted to an `int` when inserted into `generated.neighborhood_connected_census_blocks` [here](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/8a73f46c362bd926c4e316b80fc3481d551db7e3/brokenspoke_analyzer/scripts/sql/connectivity/connected_census_blocks.sql#L28).

In running analysis on other cities after those changes, `pg_stat_all_indexes` did not show any usage of the single column `base_road` / `target_road` indexes, so I removed the `target_road` indexes in favor of one multi-column index like the high stress table currently has.

With those changes, the total size of `generated.neighborhood_reachable_roads_high_stress` for Milwaukee, WI was 66GB. In running this for Madison, WI, the changes reduced the size of the table from 17GB to 11GB.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_cleanup.sql
+++ b/brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_cleanup.sql
@@ -2,8 +2,4 @@ CREATE UNIQUE INDEX IF NOT EXISTS idx_neighborhood_rchblrdshistrss_b
 ON generated.neighborhood_reachable_roads_high_stress (
     base_road, target_road
 );
-CREATE INDEX IF NOT EXISTS idx_neighborhood_rchblrdshistrss_t
-ON generated.neighborhood_reachable_roads_high_stress (
-    target_road
-);
 -- VACUUM ANALYZE generated.neighborhood_reachable_roads_high_stress;
--- a/brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_prep.sql
+++ b/brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_prep.sql
@@ -5,8 +5,7 @@
 DROP TABLE IF EXISTS generated.neighborhood_reachable_roads_high_stress;
 
 CREATE TABLE generated.neighborhood_reachable_roads_high_stress (
-    id SERIAL PRIMARY KEY,
     base_road INT,
     target_road INT,
-    total_cost FLOAT
+    total_cost INT
 );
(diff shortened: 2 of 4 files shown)


robahtou:
LGTM 🍰 👍🏽  
for reference: https://www.postgresql.org/docs/current/indexes-bitmap-scans.html

surprised they don't use the primary key. 🤷🏽‍♂️

mitchellhenke:
> LGTM 🍰 👍🏽 for reference: https://www.postgresql.org/docs/current/indexes-bitmap-scans.html
> 
> surprised they don't use the primary key. 🤷🏽‍♂️

I think in this case, it looks like the ensuing analysis is never doing a query on these tables that filters on the `id` column, so the primary index doesn't get used.

lalver1 (approved):
👍

PR #725, opened 2024-10-04
Fix SQL linter violations

rgreinho:
Fixes the violations reported by the SQL linter.


lalver1 (approved):
Great! I checked that the linter does not show any violations 👍

PR #726, opened 2024-10-04
Export all tables as GeoJSON

rgreinho:
Exports all tables as GeoJSON.

Drive-by:
- Adds the intersections tables.


lalver1 (approved):
👍

PR #727, opened 2024-10-04
Do not move tables to received schema

rgreinho:
Does not move the full OSM tables to the received schema.


lalver1 (approved):
👍

PR #729, opened 2024-10-07
Fix Dockerfile

rgreinho:
When installing the brokenspoke-analyzer from the wheel, the extras were
not being installed, causing issues between SQLAlchemy and psycopg. This
patch addresses the issue by retrieving all the runtime dependencies and
installing them before installing the brokenspoke-analyzer.


lalver1 (approved):
It works great! 👍 
I'm still playing with it on my Windows machine but it may be something unrelated to this PR.

PR #800, opened 2024-12-24
Update width calculations in width_ft.sql

mitchellhenke:
# Pull-Request

## Types of changes

## Description

This PR intends to make a couple changes around the calculation of `width_ft` from OSM widths. Both are related to the fallback case of multiplying the width by 3.28084 when no units are specified. The first tries to capture cases where the width is expressed in feet and inches using single and double quotes.

The second adds a null check to not update `width_ft` if it has already been set. I noticed that paths with an OSM width of say `8 ft` were being multiplied to become 26 ft.

Example:

```sql
SELECT width_ft, osm.width from neighborhood_ways nw
JOIN neighborhood_osm_full_line osm ON osm.osm_id = nw.osm_id;

--  width_ft | width
-- ----------+-------
--        33 | 10 ft
--        33 | 10'
--        26 | 8'
--        29 | 29'4"
--        33 | 10'
--        30 | 9'4"
--        33 | 10'
--        39 | 12 ft
--        39 | 12 ft
```

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/features/width_ft.sql
+++ b/brokenspoke_analyzer/scripts/sql/features/width_ft.sql
@@ -13,6 +13,17 @@ WHERE
     AND osm.width IS NOT NULL
     AND osm.width LIKE '% ft';
 
+-- feet specified as X'Y" OR X'
+UPDATE neighborhood_ways
+SET
+    width_ft = substring(osm.width FROM '(\d+)''')::FLOAT
+    + coalesce(substring(osm.width FROM '(\d+)"')::FLOAT / 12, 0)
+FROM neighborhood_osm_full_line AS osm
+WHERE
+    neighborhood_ways.osm_id = osm.osm_id
+    AND osm.width IS NOT NULL
+    AND (osm.width LIKE '%''%"' OR osm.width LIKE '%''');
+
 -- meters
 UPDATE neighborhood_ways
 SET width_ft = 3.28084 * substring(osm.width FROM '\d+\.?\d?\d?')::FLOAT
@@ -30,5 +41,6 @@ SET width_ft = 3.28084 * substring(osm.width FROM '\d+\.?\d?\d?')::FLOAT
 FROM neighborhood_osm_full_line AS osm
 WHERE
     neighborhood_ways.osm_id = osm.osm_id
+    AND neighborhood_ways.width_ft IS NULL
     AND osm.width IS NOT NULL
     AND substring(osm.width FROM '\d+\.?\d?\d?')::FLOAT < 20;


rgreinho:
@mitchellhenke I started to work on this PR, and I ran it on a couple of our test cities and I noticed there was no difference in execution time, or score. 

Would you please be able to clarify what the goal of this PR is and what issue it is solving? Or what feature it is adding?

mitchellhenke:
> @mitchellhenke I started to work on this PR, and I ran it on a couple of our test cities and I noticed there was no difference in execution time, or score.
> 
> Would you please be able to clarify what the goal of this PR is and what issue it is solving? Or what feature it is adding?

I'm not sure the issue would necessarily result in a significant change in time or score. The cases where it can potentially change scoring is where there is a `highway=footway` with `bicycle=yes/permissive/designated` and a width specified in feet, but less than 8 feet. The initial `width_ft` will be something like `5` if `width=5 feet`. Because the final `UPDATE` command multiplies by 3.28 to convert meters to feet without excluding the previous calculations, the `5` becomes `16`. A path that should not count becomes wide enough to be included because of the multiplication.

As a concrete example in Madison, WI, [Way 499347833](https://www.openstreetmap.org/way/499347833) should be too narrow to be included as a route in the analysis because it is tagged as `width=5 feet`, but last year's analysis appears to include it. The situations where this happens does not seem to be frequent, but there are a handful returned in this [Overpass Turbo query.](https://overpass-turbo.eu/?q=W291dDpqc29uXTsKKAoKbndyWyJoaWdod2F5Ij0iZm9vdMSYxJpdxJJ3aWR0aMSbIjQgZnQixKQiYmljeWNsZcSreWVzxLEoe3tiYm94fX0pxIrEj8SRxJPElcSXxJnEq8SexKDEosSxxKXEp8SpxKvErcSvxZXEs8S1xLfEucSrZMS9xJVuYXRlZMS_xYHFg8WFxYfFicSOxJDEksSUxJbFlMScxZLEocWQxLLEpsSoxKrEnMWaxLDEssS0xLbEuMS6xJxwZXJtaXNzaXbEusWwxYzFs8WPxJrFtsSfxbjEo8WWxbzEqzXErsaAxJLGgsWfxoUixLzEvl3FgMWCxYTFhsWIxYrFscWNxbTFkMaWxZPFucaaxZjEnMadxZvGgcWexoTFocWjZ8WlxafFqcanxavGqsWuxq3GksWOxbXEncaXxZTFusWXxb0ixrjGn8WdxoPFoMaGxojGisaMxo7EuseDxqnFrcasx53FrMarxa_Fi8Wyx4nGsceLxrPGmSLFu8a2IjbGnsWcxqHGvMScxqXFqseex6PHh8emxrDGlcepxpjFnMetx5DHsMa5xqDGu8eWIsWixo3Gv8WmxajHt8eix4bGkce7xpTFkceMxrTHrMePxKvIg8eTx7PIh8aHxonGi8aNxo_IjseFxqzIkcavyJPGsse_x47Gm8ScN8exxrrHlcajx7bHocikx6TGrsaTx4rFt8eNxrXHkMiuyITHlMaixr3IiseAyI3ItMefyLbHiMe8yJTHqsiAyJjIrcivyIXIscSryJ7Hmcihx5zGqMiPyKXFr8SBdCDFhGR5xIo-xIrJnCBza2VsIHF0Ow&c=AlGam_yG2L).

image
image

lalver1:
Thanks for the contribution @mitchellhenke , and sorry that it took me so long to review!

I agree about not seeing a big change in the scoring but this fix makes a lot of sense to me and makes the width calculation from `osm.width` more robust (cases where the width is expressed in feet and inches using single and double quotes and the null check to not update `width_ft` if it has already been set). 

I'll get feedback from @rgreinho  and @gstonecipher this Wednesday, but my feeling is to approve this PR. Thank you!

lalver1 (approved):
Looks good @mitchellhenke, thank you!

rgreinho (approved):
👍

mitchellhenke:
Thank you!

PR #801, opened 2024-12-28
Remove ways that do not allow bicycles

mitchellhenke:
# Pull-Request

## Types of changes

## Description

This PR makes a couple changes for deleting ways from OSM prior to computation of the network. This has a couple benefits in that it should both reduce incorrect routing by not calculating routes that cyclists couldn't or shouldn't take and improve performance by avoiding calculations on those routes.

The change to delete ways with the [highway=motorway](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway) tag is because the tag implies that bicycles are not allowed. The change attempts to delete any motorway that does not specify whether bicycles are allowed or that specify that bicycles are not allowed. `motorway_link` is probably not worth considering for this as at least in my quick review locally, as there are some that would be usable by bicycles.

As an example in Milwaukee, the 794 freeway would be quite convenient to reach destinations at either end and allowing it as a route likely includes destinations that would not otherwise be accessible in the allowed distance:

image

In doing a calculation with the changes proposed here, the freeway is no longer a routable path:

image

## Checklist

### DIFF
--- a/brokenspoke_analyzer/core/compute.py
+++ b/brokenspoke_analyzer/core/compute.py
@@ -58,11 +58,11 @@ def features(
     bind_params = {"nb_boundary_buffer": boundary_buffer}
     execute_sqlfile_with_substitutions(engine, sql_script, bind_params)
 
-    # Remove paths that prohibit bicycles.
-    logger.info("Removing paths that prohibit bicycles")
+    # Remove anything that prohibits bicycles.
+    logger.info("Removing anything that prohibits bicycles")
     dbcore.execute_query(
         engine,
-        "DELETE FROM neighborhood_osm_full_line WHERE bicycle='no' and highway='path';",
+        "DELETE FROM neighborhood_osm_full_line WHERE bicycle='no';",
     )
 
     # Setting values on road segments.
--- a/brokenspoke_analyzer/core/compute.py
+++ b/brokenspoke_analyzer/core/compute.py
@@ -65,6 +65,18 @@ def features(
         "DELETE FROM neighborhood_osm_full_line WHERE bicycle='no';",
     )
 
+    # Remove motorways that imply bicycles are prohibited and do not
+    # specify otherwise
+    logger.info("Removing anything that prohibits bicycles")
+    dbcore.execute_query(
+        engine,
+        """
+        DELETE FROM neighborhood_osm_full_line
+        WHERE highway = 'motorway'
+        AND bicycle IS NULL OR bicycle = 'no';
+        """,
+    )
+
     # Setting values on road segments.
     logger.info("Setting values on road segments")
     sql_scripts = ["one_way.sql", "width_ft.sql", "functional_class.sql"]


gstonecipher:
@mitchellhenke - thanks for submitting these changes - we are not going to incorporate them at this time, but I wanted to give some specific thoughts:

Regarding the motorway component: in its routing calculatings, the BNA only considers routes that exist entirely using low-stress roads and intersections. Given the speed limits and lack of bike infrastructure on motorways, those ways should always be considered high-stress, and therefore would not ever be considered in the routing, even if the road is within an allowable trip distance. In the example you gave, the 794 freeway in Milwaukee is shown as red (high stress) on the map, and so by our metrics, is not increasing bicycle connectivity or contributing to the city score. We did run some tests using your changes on a few cities, and confirmed that scores did not change.

Regarding deleting ways where bikes aren't allowed: While this concept makes a lot of sense, we are ultimately trying to show all of the road/path based infrastructure in a city, and therefore don't want to simply delete ways. This is especially important for visualization purposes on the BNA frontend map. I think there is potentially a future version of the tool that classifies ways as high stress, low stress, and not accessible (no bikes allowed), which again would make both routing and visualization more clear. However, since our tests showed no change to the scoring outputs for our sample cities, this is not highest priority at this moment. 

Thanks again for bringing our attention to this, and for your work on improving the BNA - much appreciated!

mitchellhenke:
Thanks for the consideration! I totally understand that the proposed changes aren't accepted, but I did want to double check that it would not have an impact on the score.

Apologies if my understanding is not quite correct, but in reading through the analysis it seems as though low stress consideration is in part dependent on the ratio of the distance between the high stress and low stress routes between the source and destination block. My understanding of the check [here](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/aa5afd608e212f93f2aa8d75262419ff2cf2af3e/brokenspoke_analyzer/scripts/sql/connectivity/connected_census_blocks.sql#L66) is that if the low stress route is more than 25% longer than the high stress route, it is not scored as a low stress connection. I think the [access](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/brokenspoke_analyzer/scripts/sql/connectivity/access_population.sql#L50-L54) scores also consider what proportion of the destinations are high and low stress.

If that is roughly right, the inclusion of un-bikeable routes would have an impact on scoring by potentially both creating shorter high stress routes and including destinations in the radius that would otherwise not be reachable by bike. I'd be surprised if the impact was large, but I suspect there is at least a small one.

Either way, I appreciate the consideration and understand!

gstonecipher:
Hi Mitchell - I think your understanding of the methodology is spot on, and your explanations of potential differences do make sense. I think the difference in thought may come from what defines "un-bikeable" - while excluding ways that explicitly have "bicycle=no" could change the length of the shortest high stress route, there are plenty of other high-stress routes where bicycles are technically allowed, but someone is very unlikely to ride. I think for the purposes of this analysis (especially because we're looking at scores at a city-wide level, and not trying to pick the best route for someone to actually ride), these two cases can be treated similarly. 

But again - thanks for bringing this up! This has led me to some good thinking regarding the concepts of un-bikeable due to high stress vs. un-bikeable due to legality, and I'll definitely keep thinking on it as the tool evolves.

mitchellhenke:
I appreciate the consideration. I think from my perspective, the issue goes both directions. Motorways/freeways are one aspect that results in potentially illegal high stress routes being included. On the other side, `bicycle=no` residential streets that aren't excluded can create false low stress routes that are high stress in reality (for riders or others), impacting scores in a different way.

I'm unsure if that changes the conclusion, but am grateful for the discussion. I'll plan on closing the PR.

mitchellhenke:
I've done a bit of tinkering with an approach to not delete unbikeable ways, but rather mark them as such and exclude those ways from the calculation. I've got the branch [here](https://github.com/PeopleForBikes/brokenspoke-analyzer/compare/main...mitchellhenke:brokenspoke-analyzer:mark-and-exclude-ways-that-do-not-allow-bicycles) and am happy to open up a PR for discussion review if it is of interest. Pink in the screenshot below shows ways that would be considered unbikeable.

image

PR #807, opened 2025-01-07
Use new format for street parking tagging

mitchellhenke:
# Pull-Request

## Types of changes

## Description

The structure for tagging on-street parking changed towards the end of 2022. [parking:lane](https://wiki.openstreetmap.org/wiki/Key:parking:lane) and friends was deprecated in favor of [parking:side](https://wiki.openstreetmap.org/wiki/Street_parking#To_migrate_deprecated_parking:lane=*_and_parking:condition=*_tags). The changes here attempt to capture that. I wasn't sure if the deprecated tagging should be kept or removed. It seemed like most parking tags have been updated, so I opted to not, but can re-add it if it's preferred.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/pfb.style
+++ b/brokenspoke_analyzer/scripts/pfb.style
@@ -130,6 +130,15 @@ way        parking:lane:width        text  linear
 way        parking:lane:right:width  text  linear
 way        parking:lane:left:width   text  linear
 way        parking:lane:both:width   text  linear
+way        parking:right  text  linear
+way        parking:left   text  linear
+way        parking:both   text  linear
+way        parking:right:restriction  text  linear
+way        parking:left:restriction   text  linear
+way        parking:both:restriction   text  linear
+way        parking:right:width  text  linear
+way        parking:left:width   text  linear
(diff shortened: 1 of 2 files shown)


lalver1:
Thanks for keeping up with updates to the parking tags! The changes look good to me, I am just wondering if we can re-add the deprecated tagging? I was looking at a few cities in overpass-turbo using the [osm wiki query](https://overpass-turbo.eu/?Q=%5Bout%3Ajson%5D%3B%0A(%0A%20%20way%5B~%22%5Eparking%3Alane.*%24%22~%22.%22%5D(%7B%7Bbbox%7D%7D)%3B%0A%20%20way%5B~%22%5Eparking%3Acondition.*%24%22~%22.%22%5D(%7B%7Bbbox%7D%7D)%3B%0A)%3B%0Aout%20body%3B%0A%3E%3B%0Aout%20skel%20qt%3B) and it seems like there are still some deprecated tags hanging around. Just to be safe, it might be good to keep them in the analysis. Thank you!

mitchellhenke:
Sure thing, just updated!

PR #816, opened 2025-01-16
Update ruff

rgreinho:
Updates ruff and fix CI violations.


rgreinho:
This PR is not needed anymore.

PR #817, opened 2025-01-16
Migrate to uv

rgreinho:
Migrates from `poetry` to `uv`.

We had been wanting to do so for a while and now that `poetry export` is
causing us some trouble, this is a good excuse.

The `justfile` and the `Dockerfile`, the shared GitHub workflows, the
project workflows and the documentation have been adjusted to use `uv`
as well.


rgreinho:
Obviously... We will also have to update the workflows... 🤦

rgreinho:
Alright! At this point we migrated:

- the setup ✅
- the Dockerfile ✅
- the shared GitHub workflows ✅
- the project GitHub workflows ✅

The last step is the documentation to make sure we replace all mentions of `poetry` by `uv`.

rgreinho:
And now with the documentation!

All ready for review!

lalver1 (approved):
Looks good @rgreinho!
I tested using `uv` instead of `poetry` for running the `bna` and it all worked fine in my local environment. As usual, I was not able to test the GitHub Actions workflows but reading over them, the changes made sense 👍

PR #818, opened 2025-01-20
Buffer city boundaries for clipping OSM data

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

Currently the roads imported look to be limited to the city's boundaries. It looked like previous versions of the analysis included some buffer. This PR attempts to add some buffer to the city boundaries for the import of OSM data. I'm not sure on my approach to adding the buffer, though I hope it is in roughly the right area.

I've included a before/after example of Glendale, WI, which has some more interesting boundaries where a more strict import of the boundaries can exclude destinations and routes that may be convenient. The city area is shaded pink with the different roads imported shown in the before and after using a buffer of the default distance of 2680 meters.

| Before | After |
|-----|-----|
| image | image |

## Checklist

### DIFF
--- a/brokenspoke_analyzer/cli/prepare.py
+++ b/brokenspoke_analyzer/cli/prepare.py
@@ -40,6 +40,7 @@ def all(
     block_population: common.BlockPopulation = common.DEFAULT_BLOCK_POPULATION,
     retries: common.Retries = common.DEFAULT_RETRIES,
     lodes_year: common.LODESYear = common.DEFAULT_LODES_YEAR,
+    buffer: common.Buffer = common.DEFAULT_BUFFER,
 ) -> None:
     """Prepare all the files required for an analysis."""
     # Make MyPy happy.
@@ -55,6 +56,8 @@ def all(
         raise ValueError("`retries` must be set")
     if not lodes_year:
         raise ValueError("`lodes_year` must be set")
+    if not buffer:
+        raise ValueError("`buffer` must be set")
 
     # Handles us/usa as the same country.
     if country.upper() == "US":
@@ -80,6 +83,7 @@ def all(
             block_population=block_population,
             retries=retries,
             lodes_year=lodes_year,
+            buffer=buffer,
         )
     )
 
@@ -93,6 +97,7 @@ async def prepare_(
     block_population: int,
     retries: int,
     lodes_year: int,
+    buffer: int,
     region: typing.Optional[str] = None,
 ) -> None:
     """Prepare and kicks off the analysis."""
@@ -116,7 +121,7 @@ async def prepare_(
     # Retrieve city boundaries.
     with console.status("[bold green]Querying OSM to retrieve the city boundaries..."):
         slug = retryer(
-            analysis.retrieve_city_boundaries, output_dir, country, city, region
+            analysis.retrieve_city_boundaries, output_dir, buffer, country, city, region
         )
         boundary_file = output_dir / f"{slug}.shp"
         console.log("Boundary files ready.")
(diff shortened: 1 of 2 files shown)


rgreinho:
Hi @mitchellhenke! 

The buffer was purposely removed as it created some issues, especially with contiguous cities. We're not sure we want to add it back. Besides we were planning to retrieve the city data directly from [OSMnx](https://github.com/gboeing/osmnx) in a future version, removing the need for clipping.

We're going to think about it internally and just in case there is a good reason for us to add it back and we'll get back to you. 

Did you have a use case where the buffer was needed?

mitchellhenke:
> Hi @mitchellhenke!
> 
> The buffer was purposely removed as it created some issues, especially with contiguous cities. We're not sure we want to add it back. Besides we were planning to retrieve the city data directly from [OSMnx](https://github.com/gboeing/osmnx) in a future version, removing the need for clipping.
> 
> We're going to think about it internally and just in case there is a good reason for us to add it back and we'll get back to you.
> 
> Did you have a use case where the buffer was needed?

In running some analyses, some of the more complex borders can lead to cases where blocks could have pretty different scores based on how much you include in adjacent cities.

I'm probably not recognizing anything new within this project, but I don't think there's a straightforward answer between whether it's preferable to score the infrastructure as it exists strictly within the borders of the city or scoring whether someone living in a city can reach people/jobs/destinations nearby (including those on city borders). I tend to think of it as the latter. Infrastructure that connects me to friends, etc. in a nearby municipality regardless of whether the infrastructure is in my city or theirs is helpful for me either way.

I think there is a benefit to including a small buffer to help relieve the stricter constraints of the sometimes winding borders of cities, and perhaps even larger, but I definitely understand either way.

Below are some examples for the city of Milwaukee where borders with adjacent cities gets kind of odd.

| Glendale | Wauwatosa | Greenfield |
|---|---|---|
| image | image | image |

rgreinho:
I see. Thanks a lot for the explanation. 

What you described was the original intent behind the buffer, and makes a lot of sense from a user point of view. 

However we noticed that people like city planners, and mobility/transportation advocates wanted to be more focused on their own city, so we removed it. 

We also wanted to provide a score purely based on the infrastructure development of the city itself, and not its neighbors  (e.g. avoid a scenario where city A would get +5pt score even though they did not improve anything, but just because city B next to them did a lot of good work).

We could propose to add an optional flag, or a feature flag for this, so that users could decide for themselves if they wanted to use a buffer or not. However this would also conflict with our future work of retrieving the city data directly with OSMnx and would need to be removed once this change would be implemented. It is a bit tricky and we need to think a bit more about it.

I'll revive the discussion during our working group meeting and I'll comment here once we find an agreement.

mitchellhenke:
Makes sense to me! I don't think it's necessarily worth the complexity to support the buffer, especially since it sounds like it would require keeping a separate functionality for importing outside of the city boundary.

I'm very happy to pick it back up if anything changes though.

mitchellhenke:
I ran into a kind of interesting example of the impacts of the stricter clipping. The Hank Aaron State Trail runs near the border of Milwaukee and West Allis, and while the majority of sections are exclusively in Milwaukee, some sections are in both, and one small section (pictured below) is entirely in West Allis.

I understand the reasoning to not want adjacent municipalities to benefit from improvements they didn't build, but I think there's a decent case to add _some_ buffer. A given path/street may not even be owned by a specific municipality and instead an encompassing jurisdiction like a county or state does. In this case, the Hank Aaron State Trail is owned and maintained by the State of Wisconsin.

The first image below is the section in question, [way 116791193](https://www.openstreetmap.org/way/116791193), and the second is the paths that are included in an analysis of Milwaukee. The end result is that people on either side of the bridge end up being calculated as though they couldn't reach each other on low stress ways, which ends up reducing the score for both Milwaukee and West Allis.

image

image

mitchellhenke:
Similar to the above, I found another situation where the strict boundaries has kind of unintended side effects. Part of the border between Milwaukee, WI and Shorewood, WI is on Edgewood Avenue. As is sometimes the case, the street is in both municipalities. The city boundary and where the street is mapped in OSM results in the street not being part of the network in the analysis. One block group ends up on kind of an island with a score of 9 while neighboring blocks are in the 60s and 70s.

image

Issue #649, opened 2024-07-03
Opportunities for Improving Analysis Speed


mitchellhenke:
# Help request

Hello! I've been running analysis locally to see how current projects affect the scores of places near me, but it is time-prohibitive for larger cities as it can take many hours to run (ex: Madison, WI).

## Problem

I've been digging in a bit to some of the parts that seem to take longer, and two I've noticed taking longer are:

- `brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_calc.sql`
- `brokenspoke_analyzer/scripts/sql/features/functional_class.sql`

I've ran a few dozen analysis on cities of different sizes to try to understand where the best opportunities are, but I figured I'd open an issue as you all probably have a better idea. If there's anything specific you've got, I'm more than glad to dig in and see what I can do. Thanks!

mitchellhenke:
Hi!

I had some various other attempts at improving the speed, but I never wrapped it up because I wasn't sure if it would maintain correctness or if the performance benefits worked well across different sized cities.

I've pushed it here, and am happy to pick up and work on anything that seems promising: https://github.com/PeopleForBikes/brokenspoke-analyzer/compare/main...mitchellhenke:brokenspoke-analyzer:mitchellhenke/other-performance-stuff

rgreinho:
Alright! I love the idea! 

So using your PR (#821) with threads, I ran a similar experiment using asyncio. 

I compared using a database with 1 core vs 8 cores, and the main branch, vs threads, vs asyncio. The results were a bit surprising, as there is very little difference when the database has 1 core. The threads perform better only on 8 cores, and for big cities.

On production, our ephemeral databases only have 1 core, so that's our main use case.  

The results are in the wiki: https://github.com/PeopleForBikes/brokenspoke-analyzer/wiki/async-vs-threads

I think we need to dig more to better understand the impact of these changes, but it looks like that if we want to increase the performance it will have to be done at the database level rather than at the brokenspoke-spoke analyzer level.

mitchellhenke:
Thanks for running all of those experiments!

> I think we need to dig more to better understand the impact of these changes, but it looks like that if we want to increase the performance it will have to be done at the database level rather than at the brokenspoke-spoke analyzer level.

[...]

PR #821, opened 2025-01-20
Parallelize computation of reachable roads and connected census blocks

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

This is one of the bigger changes I've been experimenting with as part of #649. It should likely be configurable, but I figured it was worth seeing whether the approach is close to the right track. I don't have a huge amount of experience with concurrency in Python, so it's been a bit of a learning curve. The results have been successful enough to me that I wanted to propose something along these lines.

I tested on smaller cities and didn't notice a significant change in either direction. Where it was noticeable was on larger cities. I used Madison, WI as my test case because it is larger, but not too large to run in less than a day on my laptop. There is likely some tuning that could be done. In Python 3.13, ThreadPoolExecutor will default to `min(32, (os.process_cpu_count() or 1) + 4)` workers. My laptop has 8 cores, so it ran with 12 workers.

| Calculation | Duration Before | Duration After | % Improvement |
|-----|-----|----|-----|
| Reachable Roads | 7 hours, 40 minutes | 3 hours, 7 minutes | 59% faster |
| Connected Census Blocks | 26 minutes | 9 minutes | 65% faster |

## Checklist


brokenspoke_analyzer/core/compute.py
```diff
@@ -22,10 +24,12 @@ def execute_sqlfile_with_substitutions(
     engine: Engine,
     sqlfile: pathlib.Path,
     bind_params: typing.Optional[typing.Mapping[str, typing.Any]] = None,
+    log_percentage: typing.Optional[float] = None,
 ) -> None:
     """Execute SQL statements with substitutions."""
-    logger.debug(f"Execute {sqlfile}")
-    logger.debug(f"{bind_params=}")
+    if log_percentage == None or uniform(0, 1) <= log_percentage:
```
mitchellhenke, line 30: The logging is extremely noisy can calculating roads and census blocks individually. I wasn't sure on the best method to log _some_ during the longer calculation steps, but ended up with this as a configurable way to be able to track progress.

brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_calc.sql
```diff
@@ -16,6 +16,7 @@ SELECT
 FROM neighborhood_ways AS r1,
     neighborhood_ways_net_vert AS v1,
     neighborhood_ways_net_vert AS v2,
+    -- WHERE clause for neighborhood_ways_net_link based on road_id?
```
mitchellhenke, line 19: `neighborhood_ways_net_link` can be pretty large, and avoiding selecting all of the vertices would probably speed up the reachable roads calculation further. An index would be likely be needed though. I experimented a bit with using the below WHERE clause an index.

```sql
WHERE ST_DWithin(l.geom, (
  select ST_Union(geom) from neighborhood_ways where road_id =1
), 3000)
```

```sql
CREATE INDEX IF NOT EXISTS tsidx_neighborhood_ways_net_link_geoms
ON neighborhood_ways_net_link USING GIST (geom);
```

rgreinho:
I like this idea very much!

Not sure how I feel about the Threadpool though, because Python is limited to more or less one core by the GIL anyway, but I really like the concept!

I am going to rewrite you PR using `asyncio`, which usually as good results in Python, and compare the results. We could see what we merge afterward.

brokenspoke_analyzer/core/compute.py
rgreinho: Yeah, not sure about this though. I think I prefer to have too much logs rather than a sample. Maybe we could just increase the `log_level` to `trace` in this case?

brokenspoke_analyzer/core/compute.py
```diff
@@ -321,13 +332,20 @@ def conectivity(
...
+                    sql_script,
+                    {"road_id": i, "nb_max_trip_distance": max_trip_distance},
+                    0.001,
+                ): i
+                for i in road_ids
             }
-            execute_sqlfile_with_substitutions(engine, sql_script, bind_params)
+            for future in concurrent.futures.as_completed(future_to_road_id):
+                url = future_to_road_id[future]
+                data = future.result()
```
rgreinho, line 349: The previous code was doing 8 iterations for calculating the stress level (although I am not sure why 8 was choosen...). Where did this go? We don't need it because we run it once per road id?

brokenspoke_analyzer/core/compute.py
```diff
@@ -303,6 +307,13 @@ def conectivity(
     }
     execute_sqlfile_with_substitutions(engine, sql_script, bind_params)
 
+    # Fetch all road_ids and then calculate stress for them individually but in parallel
+    result = dbcore.execute_query_with_result(
+        engine, "select road_id FROM neighborhood_ways"
+    )
+    road_ids = [e for l in result for e in l]
```
rgreinho, line 314: Just for info here, when you are trying to do something with iterators, and moreover when that seems to be a very common operation, the `itertools` package probably has the answer you need.

In this case I found out that `https://docs.python.org/3/library/itertools.html#itertools.chain` can be used to flatten a list of tuples directly.

brokenspoke_analyzer/core/database/dbcore.py
```diff
@@ -12,6 +12,12 @@
 from brokenspoke_analyzer.core import runner
 
 
+def execute_query_with_result(engine: Engine, query: str) -> typing.Any:
```
rgreinho, line 15: I would specify the full return type here: `typing.Sequence[Row[typing.Any]]`.

brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_calc.sql
rgreinho: Cool! What was the result of the experimentation?

brokenspoke_analyzer/core/compute.py
mitchellhenke: Yes, that makes sense!

brokenspoke_analyzer/core/compute.py
mitchellhenke: I wasn't sure on the current implementation. From reading it, it seemed like the modulo 8 meant it was effectively splitting up the entire collection of roads into 8 groups, and each of those groups was updated once. The parameter was named `:thread_num`, but as far as I could tell there were no threads, and it was updating the "groups" sequentially.

brokenspoke_analyzer/core/compute.py
mitchellhenke: Thanks for the suggestion, updated to use `itertools`!

brokenspoke_analyzer/core/database/dbcore.py
mitchellhenke: Ah, yep, I meant to follow up and fix that. Should be fixed now.

brokenspoke_analyzer/scripts/sql/connectivity/reachable_roads_high_stress_calc.sql
mitchellhenke: In some a limited test, it seemed significant. I'm not sure how much is attributable to just loading less and how much loading less helps reduce the calculation that `PGR_DRIVINGDISTANCE` has to do though.

As an example case, Wauwatosa, WI has around 41,000 rows in `neighborhood_ways_net_link`. When limiting to `:nb_max_trip_distance + 100` (which is I think a more than safe enough buffer?), the number returned for a given road can be quite a bit lower, usually less than half, sometimes much less. The runtime test I did was not terribly scientific, but in using the timestamps from calculating the high/low stress roads (not including "cleanup"), the results for Wauwatosa were:

| Calculation | Duration Before | Duration After | % Improvement |
|-----|-----|----|-----|
| High Stress | 156 seconds | 93 seconds | 40% faster |
| Low Stress | 71 seconds | 47 seconds | 33% faster |

I also ran it with/without for Milwaukee, but didn't keep track of the difference, but I want to say I recall it being significant.

At least for the case of Wauwatosa, it doesn't end up changing results in `neighborhood_connected_census_blocks`, which gives me some more confidence in it not resulting in calculation changes. I've pushed up ad83cbca3121aa2aa0eea9b43c10a3069a1af422 to this PR as the implementation I've used for the above.

brokenspoke_analyzer/core/compute.py
rgreinho: Ah yes, in the original implementation is was called in parallel. 

You could find it here: https://github.com/PeopleForBikes/pfb-network-connectivity/blob/develop/src/analysis/scripts/run_connectivity.sh#L53-L65

rgreinho:
@mitchellhenke Since we need to do more research about this, I am marking this PR as draft.

And I am putting some ideas from PostGIS here: https://postgis.net/docs/manual-3.0/performance_tips.html for reference.

PR #832, opened 2025-03-08
Fix water block issues

rgreinho:
Ensures the water blocks are deleted for US cities.

Drive-by:
- Adds utility functions to normalize country names and detect US
  aliases.


lalver1 (approved):
👍

Issue #254, opened 2023-07-29
Cache downloads


rgreinho:
# Feature request

## Current Behavior

The files are downloaded every time we delete the prepared files or when we run the tests.

## Expected Behavior

We should cache the downloads instead of re-downloading the files for nothing.

## Possible Solution

- We should create a `.bna_cache` folder which would contain the files successfully downloaded.
- When we would prepare the fhe files required for an analysis we would check in the cache folder first.
  - For example the water_census_blocks file or the speed_limits file are always the same for any US city. The OSM files should be cached as well in case we re-run an analysis for a city, or for several cities in the same state/region.

rgreinho:
When caching the downloads, we must also add the ability to disable the cache. The use case for that is that the cache may not always be necessary, for instance when using this tool in a Docker container.

rgreinho:
Another idea would be to detect the standard cache directory per platform and use it:

- macOS: ~/Library/Caches/<appname>
- Windows: C:\Users\<username>\AppData\Local\<appname>\Cache
- Unix: ~/.cache/<appname>

rgreinho:
It looks like [PlatformDirs](https://pypi.org/project/platformdirs/) would come to the rescue 😃

rgreinho:
@lalver1 I'm taking this PR as part of my efforts to keep the census files in a common place.

PR #842, opened 2025-03-25
Use cache folder for downloads

rgreinho:
Uses platform specific cache folder to download files common to all
analyses.

The cache can also be an AWS S3 bucket.

When fetching file, the downloader always looks for the file in the
cache first, then either fetches it from the internet if it does not
exist, or simply copy it from the cache to the data store.

The documentation was updated accordingly.

Since the cache really speeds up the "prepare" stage, the operations
faster than 80ms were not being displayed on screen as the `Console`
object did not have time to refresh. As a result, the output was
adjusted to always display log messages, but only show the spinner if
the operation takes longer than 80 ms, which seemd to be a reasonable
compromise.

Drive-bys:
- Updates runtime and dev dependencies.
- Adds a script to warmup the cache.
- Fixes SQL lint violations.

Fixes PeopleForBikes/brokenspoke-analyzer#254


rgreinho:
First attempt here. 

Only tested locally with some XS US cities, but it worked. We might have some issues in the cloud as is.

This will likely be a breaking change though, since the location of the input file changes.

rgreinho:
Alright! At this point the logic was implemented and tested, so we can start playing with it to ensure nothing broke. It should not be a breaking update anymore.

The documentation is missing, but will come soon.

lalver1:
This is looking good Rémy! As I'm playing with it I'm noticing a few things, I'll write them here but I'm still thinking about them:
- If we are running the analyzer in the dev container, [`get_user_cache_dir`](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/8f1e9c28a8e56fd7247df619fd12c30721863736/brokenspoke_analyzer/core/utils.py#L276) returns `/home/bna/.cache/brokenspoke-analyzer`. Do we need to map this location to the user's local cache directory using a Docker volume or something like that so that we don't lose the cached files when the dev container is deleted?
- Minor thing, we'll need to add `graphviz` to the applications `apt` needs to install in the `dev` stage of the docker image build so that the documentation can show the caching graphviz diagram. I can do this as a quick follow up PR.

rgreinho:
Good catch @lalver1 ! 

So I made sure this patch would not be a breaking change by disabling the cache by default. The main reasoning was that the containers would need a little bit of extra work to accommodate for this change.

Since we're going to make it the default behavior for v3, I think it is acceptable if we don't guarantee that the cache would work in the container. Do you agree with that statement? Otherwise if you think it is a blocker I can plan to do the necessary work in a separated PR that we will merge just after. 

I'll add the `graphviz` package to the dev container 👍

rgreinho:
@lalver1 I adjusted the Console output as per our discussion.

The cache really speeds up operations, and when files are cached, it takes only a few milliseconds to retrieve their content.

So I changed the output to always display a message, but the spinner will only appear if an operation takes longer than 80ms. This seemed like a reasonable compromise to me.

The new output looks like this when all the files are cached:

image

You could notice that the "prepare" stage completes in less than 1s in this case 😄

rgreinho:
@lalver1 I updated the base Docker image to get the latest security updates, and added `graphviz` at the same time.

rgreinho:
@lalver1 alright! I think it is now ready for the final review! 😊

lalver1 (approved):
Looks great @rgreinho! 👍

PR #845, opened 2025-04-04
Add ability to run partial analysis

rgreinho:
Adds the ability to select wich parts of the analysis to run (e.g.
features, stress or connectivity).

The compute logic was updated to account for the dependencies between
the different parts. For instance, the "stress" part requires the
"features" part to run. This determination is made automatically for the
user.

The export logic was also updated to prevent non-existing tables to be
exporting, which would result in a crash. Indeed, if not all the parts
were run, not all the tables were created.

The documentation was updated accordingly.


rgreinho:
@lalver1 It sill needs to be documented, but other than that, the feature is there an working.

lalver1 (approved):
This looks great @rgreinho!
I tested a few `--with-parts` scenarios (features and stress) and they worked fine 👍

PR #846, opened 2025-04-05
Bypass Cloudflare Anti Download Bot

rgreinho:
The census site started to use Cloudflare to block bots, therefore also
affecting us while retrieving the census data for our analyses.

This patch updates the header to simulate a request from the Firefox web
browser, and enable cookies.


rgreinho:
Should not be needed anymore. Closing for now. I'll re-open if the problem comes back.

PR #847, opened 2025-04-19
Exclude all ways with access no/private

### LINKED ISSUE #840: Add checks for `access` tags for all ways
# Bug report

Currently the `access=*` tag only gets checked for some kinds of ways in the [functional_class](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/brokenspoke_analyzer/scripts/sql/features/functional_class.sql) query. 

## Current Behavior

Ways with `access=no` are shown as low stress ways, e.g. https://www.openstreetmap.org/way/1309414435/

Image

## Expected Behavior

Any way with `access=no` or `access=private` should be excluded.

## Possible Solution

Probably adding this clause to all classes of neighborhood_ways should fix the issue

```
AND (osm.access IS NULL OR osm.access NOT IN ('no', 'private'))
```

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

This change proposes adding additional exclusions for ways that have access=no or access=private to close #840. The exclusion exists for footways currently, but not other ways. I think it's probably correct and minimally impactful to add the exclusion to every case.

In running analysis for Madison, WI after applying this patch, the private path ([1309414435](https://www.openstreetmap.org/way/1309414435/#map=18/43.041740/-89.402398)) that was part of the report in #840 is shown crudely highlighted is now not part of the network as expected:

image

## Checklist

Fixes: #840


rgreinho:
I am a bit wary of using `access=no` as I discovered this while working with OSNMX:

> Filter for the OSM “access” tag. Default is `[“access”!~”private”]`. Note that also filtering out `access=no` ways prevents including transit-only bridges (e.g., Tilikum Crossing) from appearing in drivable road network (e.g., `[“access”!~”private|no”]`). However, some drivable tollroads have `access=no` plus a `access:conditional` tag to clarify when it is accessible, so we can’t filter out all `access=no` ways by default. Best to be permissive here then remove complicated combinations of tags programmatically after the full graph is downloaded and constructed.

_source: <https://osmnx.readthedocs.io/en/stable/user-reference.html#osmnx.graph.graph_from_place>, for the `default_access: str` argument_

Shall we just filter with `access=private` as well? This sounds reasonable to me. But maybe finding a slightly more complex logic, as OSMNX recommends, may provides better results? 🤔

vgXhc:
Seems like the Tillikum crossing example isn't really relevant. The bike paths are mapped [as separate ways](https://www.openstreetmap.org/way/389738959), and the transit portion of the bridge _shouldn't_ show up in the "drivable road network" nor in the BNA. I'm not sure about the toll roads.

There's probably more merit to account for cases when people tag `access=no` (or private) and then add `bicycle=yes` (or permissive/dedicated/whatever) to tag a bike-only or bike/ped-only way. This is [not uncommon](https://taginfo.openstreetmap.org/tags/access=no#combinations) So filtering only by `access=no` is probably too coarse. It'd be something like `access=no AND !bicycle=(yes|permissive|designated)`?

mitchellhenke:
Based on the above, I've added a commit so that the following ways are included:

- `access` is not defined
- `access` is not `private` or `no`
- `access` is `no` and `bicycle=(yes|permissive|designated)`

As an [overpass query](https://overpass-turbo.eu/?Q=%5Bout%3Ajson%5D%5Btimeout%3A25%5D%5Bbbox%3A%7B%7Bbbox%7D%7D%5D%3B%0Away%5B%22access%22%3D%22no%22%5D%5B%22bicycle%22%7E%22permitted%7Cdesignated%7Cyes%22%5D-%3E.bicycle_connections%3B%0Away%5B%21%22access%22%5D-%3E.bicycle_connections%3B%0Away%5B%22access%22%21%3D%22private%22%5D-%3E.bicycle_connections%3B%0A%0A.bicycle_connections%20out%20geom%3B&C=43.033043%3B-87.9029%3B17), it would be something like:

```
way[!"access"];
way["access"!="private"];
way["access"="no"]["bicycle"~"permitted|designated|yes"];
```

The [inverse query](https://overpass-turbo.eu/?Q=%5Bout%3Ajson%5D%5Btimeout%3A25%5D%5Bbbox%3A%7B%7Bbbox%7D%7D%5D%3B%0Away%5B%22access%22%3D%22no%22%5D%5B%22bicycle%22%7E%22permitted%7Cdesignated%7Cyes%22%5D-%3E.bicycle_connections%3B%0Away%5B%21%22access%22%5D-%3E.bicycle_connections%3B%0Away%5B%22access%22%21%3D%22private%22%5D-%3E.bicycle_connections%3B%0A%0A.bicycle_connections%20out%20geom%3B&C=43.033043%3B-87.9029%3B17) of ways that would be excluded:

```
way["access"="private"];
way["access"="no"]["bicycle"!~"permitted|designated|yes"];
```

vgXhc:
I think the overpass link for the inverse query got messed up. Here's what I think is the correct one. https://overpass-turbo.eu/s/23po

I took a quick look at what would be excluded in my area and it correctly removes the way that prompted the initial bug report. Other removals all seem to be correct -- I didn't see many false positives, and those were based on incorrect/dubious tagging. Most of the true negatives probably would have previously made it into the analysis as low stress -- stuff like `highway=service` `service=driveway` with an explicit `access=no` and the like. So I think this is good!

lalver1:
We went over these changes and they make a lot of sense, thanks! We'll approve this PR soon to get it merged, we just want to make sure that the documentation on our end is updated (if needed) and that this change does not conflict with anything else in the methodology.

mitchellhenke:
Thank you! If there's anything else I can update for this, just let me know.

lalver1 (approved):
👍

PR #850, opened 2025-04-23
Use 2020 Census Population and Employment Data

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

I happened to be working with LODES employment data on another project, and so I started looking at the Census data used here and saw it was using older data. It's a relatively significant change to try to make, so I've done my best to minimize the diff. I tested a couple of smaller cities and it seems to work as expected so far, but I'm not sure how confident I am in the changes at this point.

### Population Data

The `TIGER2010BLKPOPHU` dataset does not seem to exist for 2020, so I've switched to data that I think is a suitable replacement. The water census blocks seems to be maintained outside of the project, so I've replaced it with using the land area column (`aland20`) in the data to remove blocks without any land area. Most of the changes are changing from `10` to `20`. The new dataset also uses `geoid` instead of `blockid`, so I opted to rename the column to maintain consistency.

### Employment Data

LODES8 is based on 2020 Census geographies, but the data structure is otherwise the same, so the changes here are smaller.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning].
 
 ## [Unreleased]
 
+## [2.6.2] - 2025-07-23
+
+### Fixed
+
+- Excluded golf courses paths from bicycle infrastructure. [#889]
+
+[#889]: https://github.com/PeopleForBikes/brokenspoke-analyzer/pull/870
+[2.6.2]:
+  https://github.com/PeopleForBikes/brokenspoke-analyzer/releases/tag/2.6.2
+
 ## [2.6.1] - 2025-06-29
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ docker run \
   --rm \
   --network brokenspoke-analyzer_default \
   -e DATABASE_URL \
-  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.1 \
+  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.2 \
   -vv configure custom 4 4096 postgres
 ```
 
@@ -84,7 +84,7 @@ docker run \
   --rm \
   --network brokenspoke-analyzer_default \
   -e DATABASE_URL \
-  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.1 \
+  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.2 \
   -vv run "united states" "santa rosa" "new mexico" 3570670
 ```
 
@@ -97,7 +97,7 @@ docker run \
   -u $(id -u):$(id -g) \
   -v ./results:/usr/src/app/results \
   -e DATABASE_URL \
-  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.1 \
+  ghcr.io/peopleforbikes/brokenspoke-analyzer:2.6.2 \
   -vv export local "united states" "santa rosa" "new mexico"
 ```
 
--- a/brokenspoke_analyzer/scripts/pfb.style
+++ b/brokenspoke_analyzer/scripts/pfb.style
@@ -106,6 +106,8 @@ way        cycleway:both:width  text linear
 node,way   flashing_lights      text linear
 node,way   foot         text         linear
 way        footway      text         linear
+way        golf         text         linear
+way        golf_cart    text         linear
 node,way   healthcare   text         polygon
 node,way   highway      text         linear
 node,way   junction     text         linear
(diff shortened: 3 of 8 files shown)


rgreinho:
Thanks a lot for this PR @mitchellhenke! This was definitely something on our radar! 

However since it is a breaking change, we planned it for v3.0.0. So I am assigning the "do-not-merge" label to the PR right to prevent the bot from closing it.

I'm going to review a potentially make some comments, but as a couple PRs will be merged first, don't feel pressured to address them yet. There will be a couple rebases in between.

rgreinho:
> The new dataset also uses geoid instead of blockid, so I opted to rename the column to maintain consistency.

Since we plan on making some cleanup changes, like removing the "neighborhood" prefixes, I think that we can use the `geoid` identifier, as it is specified in the dataset.

brokenspoke_analyzer/core/analysis.py
```diff
@@ -195,7 +195,7 @@ def create_synthetic_population(
     grid = gpd.GeoDataFrame(
         {
             "geometry": cells,
-            "POP10": population,
+            "POP20": population,
             "BLOCKID10": [
```
rgreinho, line 199: I am assuming this should become `BLOCKID20`.

brokenspoke_analyzer/core/ingestor.py
```diff
@@ -123,18 +123,14 @@ def load_water_blocks(engine: Engine, csvfile: pathlib.Path) -> None:
 
 def delete_water_blocks(engine: Engine) -> None:
     """Delete the water blocks located within the city boundaries."""
-    query = (
-        f"DELETE FROM {CENSUS_BLOCKS_TABLE} AS blocks "
-        f"USING {WATER_BLOCKS_TABLE} AS water "
-        "WHERE blocks.BLOCKID10 = water.geoid;"
-    )
+    query = f"DELETE FROM {CENSUS_BLOCKS_TABLE} AS blocks WHERE blocks.aland20 = 0;"
```
rgreinho, line 126: I like that very much! This would allow us to stop generating the waterblocks manually. We are going to double check that it produces identical or similar results, but otherwise, I like it.

brokenspoke_analyzer/core/ingestor.py
```diff
@@ -183,6 +179,11 @@ def import_neighborhood(
         load_water_blocks(engine, water_blocks_file)
         delete_water_blocks(engine)
 
+    transform_query = (
+        f"ALTER TABLE {CENSUS_BLOCKS_TABLE} RENAME COLUMN geoid20 TO blockid20"
+    )
+    dbcore.execute_query(engine, transform_query)
```
rgreinho, line 185: As mentioned previously, this will likely not be needed.

rgreinho (commented):
All of it makes sense. We will run the changes through our list of test cities to ensure they make sense, but I don't anticipate there will be any problem.

brokenspoke_analyzer/core/ingestor.py
mitchellhenke: I forgot to double check this, but it appears at least for Wisconsin, there are a very small percentage of (24 out of 9,680) blocks where `aland20` is 0, but the population is greater than zero. Would it be preferable to keep the populated blocks?

brokenspoke_analyzer/core/analysis.py
mitchellhenke: Thanks, fixed!

brokenspoke_analyzer/core/ingestor.py
rgreinho: We'll check how we generate them, and we'll get back to you. We could then compare the two methods.

brokenspoke_analyzer/core/ingestor.py
mitchellhenke: Sounds good! Deleted this and switched the `blockid20`s to `geoid20`s for this table. I wasn't sure how to handle the other tables, so I've left them as `blockid20`.

brokenspoke_analyzer/core/ingestor.py
rgreinho: Alright, so we do the exact same thing, but for some historical reasons we do it in a separate process... 🤔 Therefore we can totally keep what you did, it will consolidate and simplify the process.

We don't need to account for the population when `aland20` is 0.

mitchellhenke: Sounds good, will leave it as is!

rgreinho: To cleanup a bit more, could you remove the `def load_water_blocks()` and its references, as well as the `WATER_BLOCKS_TABLE` constant please?

mitchellhenke: Yep, I think I got all of them

rgreinho:
@mitchellhenke I am going to merge this change in our v3 branch. That's why I am editing your PR a little bit.

PR #851, opened 2025-04-24
Normalize country name before building location slug in compute command

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

I'll often run commands individually, `bna import all`, `bna compute`, etc. I recently ran into an issue with compute sometimes not finding the right boundary file when running something like `bna compute us monona wisconsin` because it would be looking for `monona-wisconsin-us.shp` and not find it because the `prepare` command uses the normalized version of country (and saved it as `monona-wisconsin-united-states.shp`). It seems like the boundary file should always use the normalized version of the country based on other call [here](https://github.com/peopleforbikes/brokenspoke-analyzer/blob/8ccec40fe20f738961a90675ffa531d8558e8250/brokenspoke_analyzer/cli/prepare.py#L60) at the beginning of the `prepare` command.

This PR moves the country normalization call above the building of the slug used to find the boundary file path, similar to what is done [here](https://github.com/peopleforbikes/brokenspoke-analyzer/blob/8ccec40fe20f738961a90675ffa531d8558e8250/brokenspoke_analyzer/core/ingestor.py#L525-L535).

## Checklist

### DIFF
--- a/brokenspoke_analyzer/cli/root.py
+++ b/brokenspoke_analyzer/cli/root.py
@@ -124,6 +124,7 @@ def compute_cmd(
     engine = dbcore.create_psycopg_engine(database_url)
 
     # Prepare directories.
+    country = utils.normalize_country_name(country)
     _, slug = analysis.osmnx_query(country, city, region)
     traversable = resources.files("brokenspoke_analyzer.scripts.sql")
     res = pathlib.Path(traversable._paths[0])  # type: ignore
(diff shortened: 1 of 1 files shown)


rgreinho (approved):
Good catch! 👍 

Thank you for fixing the bug 🐞

PR #854, opened 2025-05-03
Create a script to run a series of analysis.

rgreinho:
Creates a script reading a series of cities to process from a CSV files,
and passing them all to the `bna run-with compose` command.

Drive-bys:
- Fixes default values.
- Fixes console output and re-adjusts colors.
- Adjusts the docker compose output according to the log level.
- Adjusts the spinners and logging messages.


rgreinho:
The output looks like this, for processing 2 cities:

image

rgreinho:
@lalver1 I added a sort-of-cache-mechanism to the script to prevent downloading the Geofabrik data too many times.

rgreinho:
Grace found couple issues:
- Downloading files into cache fails often as there are no retries
- Multiple files where copied from the cache into the data folder

rgreinho:
I added the retry mechanism, but I was not able to reproduce the 2 other bugs.

PR #856, opened 2025-05-03
Fix calver revision

rgreinho:
The `calver_revision()` function was not accounting for multiple digit
revisions, therefore was replacing existing files after the 10th run
within the same month.


lalver1 (approved):
👍

PR #860, opened 2025-05-06
Fix measure

### LINKED ISSUE #858: `measure` not aggregating mileage by infra type
# Bug report

The `measure` compute command part does not seem to be aggregating the `TF` and `FT` infra values by infra type.

## Current Behavior

The mileage table displays the same infra type twice.

## Expected Behavior

The mileage table should display an infra type only once.

## Steps to Reproduce

Run the bna on Provincetown, MA, to see, for example, `sharrow` appear twice in the `mileage` table.

lalver1:
Closes #858 

Before, mileage was calculated and reported in a directional way, so there was a mileage for each of `ft_` and `tf_`. This PR changes the `measure` calculation so that mileage is added up for both directions. A LATERAL join is used to avoid scanning the same table twice.

### DIFF
--- a/brokenspoke_analyzer/cli/root.py
+++ b/brokenspoke_analyzer/cli/root.py
@@ -151,6 +151,7 @@ def compute_cmd(
             state_default_speed=state_default_speed,
             city_default_speed=city_default_speed,
             import_jobs=import_jobs,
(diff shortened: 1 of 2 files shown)


rgreinho (approved):
👍

PR #874, opened 2025-06-11
Fix bikeable service road classification

lalver1:
- **Add off-street paths to measure calculation**
- **Set the functional_class of bikeable service roads to unclassified**


rgreinho:
Per our conversation with @gstonecipher, you should be good to rebase and we can merge it @lalver1 🎉

PR #888, opened 2025-07-12
Remove redundant LODES table updates

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

The alteration of the column types and padding of the column values can be quite slow at around 6-10 seconds for Wisconsin (so there is maybe some overlap with #649) and seems to be redundant in my interaction with the LODES data. I'd expect it to be a pretty significant error if the source data had a census block identifier that was the incorrect length, and I'm not sure trying to pad the data would always result in the correct result. The inclusion of these updates seems to date pretty far back, and I couldn't find much beyond the SQL comment related to processing the imported data. Given that with the performance impact, I think removing the lines should be considered.

It might be better to add something like a check constraint on the length of the column if data validation is desired here, but there aren't any currently for other Census block columns so I wasn't sure.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/connectivity/census_block_jobs.sql
+++ b/brokenspoke_analyzer/scripts/sql/connectivity/census_block_jobs.sql
@@ -11,12 +11,6 @@
 --     as per https://lehd.ces.census.gov/doc/help/onthemap/LODESTechDoc.pdf
 ----------------------------------------
 
--- process imported tables
-ALTER TABLE state_od_aux_jt00 ALTER COLUMN w_geocode TYPE VARCHAR(15);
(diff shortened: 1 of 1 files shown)


rgreinho:
Rebased on top of `v3-base` branch.

rgreinho:
@lalver1 Would you be able to have a quick look?

lalver1 (approved):
This looks good to me @mitchellhenke, thanks!

I tested a sample city and sure enough, there were no changes to the scores. Just to make sure, I also looked at all 2022 `od_main_JT00` and `od_aux_JT00` LODES files for all states (except AK, MI, and MS because they had no 2022 data) and the `w_geocode` column always had length 15. So yep, these table updates are redundant for the latest data.

PR #889, opened 2025-07-13
Exclude golf course paths

### LINKED ISSUE #887: Golf courses included as off-street paths
Golf courses typically have highway=path, and thus are being included as off-street paths within the measure tool. The desired behavior is for golf course paths to be counted as low-stress roads, but not as off-street paths, since they aren't specifically bike infrastructure. In cases where golf=cartpath, paths should be defined as 'unclassified', similar to service roads.

lalver1:
Golf course paths will still show up as low stress roads, but not as paths since this wasn't infrastructure that was built
for exclusive biking/pedestrian use.

Fixes #887

### DIFF
--- a/brokenspoke_analyzer/scripts/pfb.style
+++ b/brokenspoke_analyzer/scripts/pfb.style
@@ -106,6 +106,8 @@ way        cycleway:both:width  text linear
 node,way   flashing_lights      text linear
 node,way   foot         text         linear
 way        footway      text         linear
+way        golf         text         linear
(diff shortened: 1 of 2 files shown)


lalver1:
Hi @gstonecipher, I tested this locally running `measure` in Austin and the mileage for off-street paths dropped from 541 (before this change) to 475. I also plotted the ways in QGIS and I can see that golf course paths are now labeled as `unclassified`. When you get a chance let me know if this change looks good to you for the are you were testing, thanks!

PR #890, opened 2025-07-17
Support new contraflow bike lane tagging

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

Using [cycleway=opposite_lane](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dopposite_lane) (or `opposite_track`) is deprecated as of last year, with [cycleway:left/right:oneway=*](https://wiki.openstreetmap.org/wiki/Key:cycleway:left/right:oneway) usage being being preferred for contraflow lanes when paired with `cycleway:left/right=*`:

> To indicate contra-flow, use `cycleway:right:oneway=-1` For bidirectional flow on that side, use `cycleway:right:oneway=no`.

The additions mostly follow the existing cases for `cycleway=opposite_lane`, and I think it's correct in that the opposite direction is set (e.g. `ft_bike_infra` is set to have the corresponding infra if `one_way_car = 'tf'` and vice versa).

I've also tested with Madison, WI which has seen a handful of ways updated to the new tagging. [way 915630711](https://www.openstreetmap.org/way/915630711#map=19/43.074234/-89.381355) and [way 157654336](https://www.openstreetmap.org/way/157654336#map=19/43.074234/-89.381355) are two interesting examples because they differ in having `oneway:bicycle=no`.

When run on `main`, I think there are a few bugs:

- Without `oneway:bicycle=no`, the reverse `tf` direction is categorized as not navigable and the forward `ft` direction infrastructure is assigned `lane` when that is the other direction
- With `oneway:bicycle=no`, the reverse `tf` direction is categorized as high stress and the forward `ft` direction infrastructure is assigned `lane` when that is the other direction

> select w.osm_id, o."oneway:bicycle", w.tf_seg_stress, w.ft_seg_stress, w.tf_bike_infra, w.ft_bike_infra from neighborhood_osm_full_line o JOIN neighborhood_ways w ON w.osm_id = o.osm_id where o."cycleway:left:oneway" IS NOT NULL AND w.osm_id IN ('915630711', '157654336') ;

  osm_id   | oneway:bicycle | tf_seg_stress | ft_seg_stress | tf_bike_infra | ft_bike_infra
-----------|----------------|---------------|---------------|---------------|---------------
 157654336 |                |               |             1 |               | lane
 157654336 |                |               |             1 |               | lane
 915630711 | no             |             3 |             1 |               | lane
 915630711 | no             |             3 |             1 |               | lane

With the changes I've proposed in this PR, the outcome is closer to what I'd expect:

  osm_id   | oneway:bicycle | tf_seg_stress | ft_seg_stress | tf_bike_infra | ft_bike_infra
-----------|----------------|---------------|---------------|---------------|---------------
 157654336 |                |             1 |             1 | lane          | sharrow
 157654336 |                |             1 |             1 | lane          | sharrow
 915630711 | no             |             1 |             1 | lane          | sharrow
 915630711 | no             |             1 |             1 | lane          | sharrow

The addition of the many negative checks is unfortunately necessary to ensure things are assigned correctly. The deprecated tagging is less popular than the new one, but is not uncommon, so I've left it in for now.

## Checklist


rgreinho:
@mitchellhenke, we will have a similar effort to update the tags for v3, so I am added this PR to the v3 milestone.

mitchellhenke:
> I am just not sure about the change to `class_adjustments.sql`. Should the condition be left as an `OR` or do we actually need to change it to an `AND`? Changing it to `AND` would likely assign fewer records (or at most the same number) to `tertiary`, since both `tf_bike_infra` and `ft_bike_infra` would have to meet the condition instead of just one.

Apologies, yeah, I could have been more specific. I noticed one-way streets with one vehicle (shared) lane and a contraflow bike lane are currently updated to "tertiary", which most of the time means the shared lane becomes high stress. This didn't seem to be the intention, but I can remove it if I've misunderstood!

PR #894, opened 2025-07-27
BNA cache

rgreinho:
This PR updates the caching mechanism and introduces a new CLI command to manage
the cache. The cache changes are reflected in the `bna prepare` command.

## New caching mechanism

In version 2 of the BNA, a caching mechanism was introduced. This mechanism
assumed that a dedicated local folder, or remote storage could be used to store
download from the census.gov websites, and would get populated as needed.

In version 3, the cache is ALWAYS a local folder, and is enabled by default. The
user has the possibility to use a flag to disable it.

The ability to be using a remote storage as cache was removed, in favor of a new
feature allowing to fetch the data from a dedicated mirror.

## BNA Cache command

A new command was added: `bna cache`.

### Usage

```bash
bna cache [OPTIONS] <COMMAND>
```

### Commands

- `bna cache clean`
  - Clear the cache, deleting all files and folders inside of the cache folder.
- `bna cache dir`
  - Show the cache directory.

## BNA Prepare command

The internals of the `bna prepare` command were changed to use the cache by
default, and to support using a mirror instead of the census websites.

* `--mirror <_mirror_url_>`
    * Specify the URL of the mirror to use to fetch the files. 

## New module

A new module, `file_utils` was created to contain the types and functions that
were used to build the `bna cache` command.

## Future work items

- The command documentation needs to be written/updated. Since there is an effort to
  rewrite the documentation following the [Diátaxis] framework, this item was
  left for later.
- The documentation for the process to create a mirror.

[Diátaxis]: https://diataxis.fr/


rgreinho:
Alright @lalver1, this PR is also ready for review 🙃

Issue #886, opened 2025-07-09
City Boundaries: mismatch between OSM and brokenspoke-analyzer


gstonecipher:
When the brokenspoke-analyzer pulls a city boundary from OSM, it should clip out all census blocks outside of the city boundaries. However, the clipped census block file (e.g., "richardson-texas-united-states.clipped.osm") still doesn't match the boundary shown on OSM.

Clipped shapefile:

<img width="453" height="292" alt="Image" src="https://github.com/user-attachments/assets/bb21f903-1e3b-4e48-add4-faf08057313d" />

OSM boundary:

<img width="500" height="349" alt="Image" src="https://github.com/user-attachments/assets/838b1e86-85e7-4de0-956c-4e08f221d98f" />

vgXhc:
Makes sense. FWIW, there is a longstanding debate whether admin boundaries belong in OSM or not (not verifiable on the ground; very easy to mess up by editors; difficult to maintain)

PR #896, opened 2025-07-30
Exclude calculating census blocks that are outside of boundary

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

While running calculations for cities, I noticed that census blocks further outside of the city boundary were being included. This might intersect with #886, but I think it may be different. Currently, the calculation deletes census blocks that are outside of a buffered area of the boundary, which ends up including census blocks in other cities. As an example, this shows the census blocks included when scoring Shorewood, WI before and after the proposed change:

| Before | After |
|---------|------|
| image  |  image  |

Census blocks do not necessarily align cleanly with municipal boundaries, so how to handle blocks on the boundary is less clear to me. I had a few ideas I considered for calculating inclusion:
- Blocks that are wholly contained within the boundary
- Blocks that intersect with the boundary
- Blocks that are at least X% within the boundary

I chose to propose including any blocks where at least 50% of the area is in the boundar. Assuming no municipal boundaries overlap, this means a given block will be included at most once across all cities. The first two options could lead to more blocks that are included in multiple cities or included in no cities. Based on discussion in https://github.com/PeopleForBikes/brokenspoke-analyzer/pull/818, my hunch was it was preferable to be somewhat conservative in what blocks are included, but implementation of any of the options is pretty straightforward.

## Checklist


rgreinho:
Seems relate d #886 as well. Marking this PR for v3.

rgreinho:
@mitchellhenke if you want you can rebase your work on the `v3-base` branch. Otherwise I'll do it when we get closer to merging the PR.

mitchellhenke:
@rgreinho done!

Issue #897, opened 2025-08-01
CLI + Documentation


rgreinho:
The CLI and its documentation need to be refreshed.

Here are some ideas about how to do so:

- Refactor the `cli` package to ensure we use 1 modules per sub-command
- Review and simplify the sub-commands whenever possible/necessary
    - e.g. `bna prepare all` could become `bna prepare` since we always prepare all the files required to run an analysis during this step.
- Review the flags and remove the unused ones
- Alphabetize the sub-commands
- Use `uv` as a model for the design/organization of the sub-commands as well as for the documentation.

Issue #866, opened 2025-05-26
Clarify the `run` command on the help page


rgreinho:
# Bug report

## Current Behavior

On the main CLI help page, the `compute` and the `run` commands have the exact same description which is confusing for the users.

![Image](https://github.com/user-attachments/assets/75a57bf3-2a1c-499a-b4ec-4bbdbf27b010)

## Expected Behavior

Write a clearer description of these 2 commands to ensure the users don't need to refer to the reference page every single time they need to remember what the `run` and `compute` commands do.

rgreinho:
Fixed in #899.

PR #899, opened 2025-08-02
Improve CLI and documentation

rgreinho:
Improves the CLI and its documentation.

Fixes PeopleForBikes/brokenspoke-analyzer#866
Fixes PeopleForBikes/brokenspoke-analyzer#897

### DIFF
--- a/brokenspoke_analyzer/cli/run_with.py
+++ b/brokenspoke_analyzer/cli/run_with.py
@@ -228,7 +228,7 @@ def run_(
     """Run an analysis."""
     # Make mypy happy.
     if not data_dir:
-        raise ValueError("`output_dir` must be set")
+        raise ValueError("`data_dir` must be set")
     if not block_size:
         raise ValueError("`block_size` must be set")
     if not block_population:


rgreinho:
Alright, this is up for review @lalver1 :)

It seems like a big PR, but it is mostly shifting things around and refactoring a little to make the CLI more robust. The "Commands" page in the documentation was almost completely rewritten though.

Here is the details of the changes:

- One sub-command per file for more clarity
- Remove unnecessary commands, like `prepare all`, which is now simply `prepare`
- Alphabetize the commands and their arguments
- `--input-dir` and `--output-dir` options were confusing. They got renamed to `--data-dir` for clarity
- Force the use of keyword arguments for some functions to prevent argument mix-up
- And the biggest part, the "Commands" documenttion page was updated, using the `uv` cli [documentation](https://docs.astral.sh/uv/reference/cli/) as a model.

rgreinho:
I also removed "Measure` from under the experimental feature flag. Since @gstonecipher will run it for all the 2025 cities, it should provide us enough feedback to give us confidence that the feature is stable.

lalver1 (changes requested):
Very nice cleanup @rgreinho!

I just have a few suggestions in the comments, and a couple more below (just for consistency, the code runs fine without these changes) since these files were not changed in this PR:

- `analysis.py`
  - in `def prepare_city_file`, the first parameter can be changed to `data_dir`. See the call to it in `analysis.prepare_city_file(data_dir, region_file_path, polygon_file, pfb_osm_file)`
  - in `def retrieve_region_file`, the second parameter can be changed to `data_dir`. See the call to it in `analysis.retrieve_region_file, region, data_dir`
- `runner.py`
  - in `def run_analysis`, the fifth parameter can be changed to `data_dir`. See the call to it in `runner.run_analysis(
            state_abbrev,
            state_fips,
            city_shp,
            pfb_osm_file,
            data_dir,
            docker_image,
            container_name,
            city_fips=city_fips,
        )`

rgreinho:
Great catch @lalver1 ! I merged all your suggestions.

rgreinho:
@lalver1 I did another pass, and I think it looks solid, therefore since you kind of approved it earlier, I am going to merge it to use it as a base for the 2 other issues I am working on that require the CLI refactor.

If you notice something you missed, we could open an issue and keep iterating.

PR #925, opened 2025-08-30
Improve the usage of the dev container

lalver1:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

This PR adds a VS Code debugger configuration to allow the user to debug the BNA from a development container.

To streamline the usage of the development container, it also adds a function to reset the database (required to do before the BNA can be ran again) from the development container. This improves the usage of the development container since the user is not required to leave the development container's terminal to reset the postgres container. Instead, the user can just use the CLI from the development container to reset the database's tables.

Currently the command is implemented as

```bash
bna configure reset [OPTIONS] PGUSER
```

in an attempt to keep it as a subcommand of `configure`, but it does not read very well.

We could also do something like

```bash
bna reset-database [OPTIONS] PGUSER
```

similar to `run-with` since I don't expect us to "reset" anything else. If we expect other things to be reset, we could do `bna reset database [OPTIONS] PGUSER`.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/cli/configure.py
+++ b/brokenspoke_analyzer/cli/configure.py
@@ -57,3 +57,11 @@ def schemas(database_url: common.DatabaseURL, pguser: PGUser) -> None:
     console.log("[green]Configuring the schemas...")
     engine = dbcore.create_psycopg_engine(database_url)
     dbcore.configure_schemas(engine, pguser)
+
+
+@app.command()
+def reset(database_url: common.DatabaseURL) -> None:
+    """Reset the database tables created by a BNA run."""
+    console.log("[green]Resetting the database tables...")
+    engine = dbcore.create_psycopg_engine(database_url)
+    dbcore.reset_tables(engine)
--- a/docs/source/commands.md
+++ b/docs/source/commands.md
@@ -56,6 +56,14 @@ The parameters are:
 - the amount of memory to allocate, in MB
 - the name of the PostgreSQL user to connect as
 
+### Reset
+
+The reset comand is a convenience command that resets the database. It deletes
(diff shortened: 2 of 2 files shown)


rgreinho (changes requested):
This is awesome @lalver1 !

Do you want to have it for a 2.6.x version (we would release 2.6.4 then)? Or should it be for v3 (in which case this PR should be rebased on top of the `v3-base` branch)?

Once it will be part of v3, the `configure` command documentation should be updated accordingly (like ]https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/v3-base/docs/source/commands.md?plain=1#L24). But this could be a separate PR.

rgreinho:
Oh, I am just thinking about it now, but we discussed the need for a custom cache directory for the dev container. I have a PR ready for review here: https://github.com/PeopleForBikes/brokenspoke-analyzer/pull/921, but this would work only for v3. So maybe it could be another PR 🤔

lalver1:
Thanks @rgreinho! I went ahead and made the changes you suggested and included the update to the docs since it'll be nice to just have it all in the same PR. But let me know if I need to make any other changes!

That's a good point about release 2.6.4 or waiting on v3 🤔 I don't have a preference, the only thing I'm wondering is if Grace would find it useful to have this feature now or if it can wait till v3. If she does think it's useful, could we merge to `main` to make it available? I guess we wouldn't need a release now since it's not a command meant for using on AWS runs that use the image, for example, it's only really used for local development.

But we can do whichever way is cleanest, I know it would make sense to have a release (even a minor one) for this since it involves changes to the CLI 🙂

Issue #929, opened 2025-09-02
City Boundaries: Pulling point instead of polygon from OSM


gstonecipher:
When pulling a city boundary from OSM, if a point also exists for that given city, the analyzer is pulling the point instead of the boundary polygon. This causes the analyzer to fail, since there is no area for which to download the rest of the data. The failure message looks like:

<img width="566" height="34" alt="Image" src="https://github.com/user-attachments/assets/458c61d6-051c-4aa1-8f62-d1832a517632" />

On OSM, it appears as if the point is part of the polygon:

<img width="806" height="619" alt="Image" src="https://github.com/user-attachments/assets/db5a3eca-2d31-4304-96b7-0623c2e770aa" />

However, this is true of every city, even those that do successfully run in the analyzer:

<img width="808" height="841" alt="Image" src="https://github.com/user-attachments/assets/24a6e332-9172-4d9d-85e5-fc7b24ca2cde" />

The issue with Claymont appears to be that the Nominatum results turn up both the census boundary (polygon) and the village (point), and the village point is being selected for download.

Another case where this occurs is Land o' Lakes, FL. Again, on OSM, there's is a point for the village and a polygon for the admin boundary. In this case, there is not point shown within this admin boundary, confirming that it's just the point for the village being pulled, and not the census boundary:

<img width="805" height="845" alt="Image" src="https://github.com/user-attachments/assets/dd1bcd10-1e9d-43c5-8742-eb09530cadce" />

Is there a way to filter by geometry type?

There's also another set of cities for which there are only points, and no polygons. For those, it would be helpful to have an error message that makes this clear, and would prompt the user to add the admin boundary polygon into OSM.

rgreinho:
I started to work on this issue, and the first thing I found while trying to reproduce the error cases you posted above is that you get the boundaries only if you use the "string query"! However we recently switched to using structured queries to fix another bug (see https://github.com/PeopleForBikes/brokenspoke-analyzer/pull/913).

If you use the structured query, Nominatim only retrieves the "village" entity which is basically a point representing the city:

- <https://nominatim.openstreetmap.org/search?format=json&polygon_geojson=1&dedupe=0&limit=1&city=claymont&state=delaware>
- <https://nominatim.openstreetmap.org/search?format=json&polygon_geojson=1&dedupe=0&limit=1&city=claymont&state=delaware>
[...]

PR #947, opened 2025-09-30
Update logic to retrieve city boundaries

rgreinho:
Uses the nominatim structured query first, then fallback to free form
query if there is no result.

Fixes PeopleForBikes/brokenspoke-analyzer#929


gstonecipher (approved):
👍

Issue #938, opened 2025-09-17
Measure Tool Bug: Georgia, USA


gstonecipher:
Instead of Georgia the state, OSM files (georgia-latest.osm.pfb) are being pulled for Georgia the country. This is causing all cities in Georgia to fail with the error:

subprocess.CalledProcessError: Command '['pgsql2shp', '-u', 'postgres', '-P', 'postgres', '-h', 'localhost', '-p', '5432', '-f', '/Users/gracestonecipher/brokenspoke-analyzer/results/united states/georgia/peachtree city/25.09/neighborhood_ways.shp', 'postgres', 'neighborhood_ways']' returned non-zero exit status 1.

The neighborhood_ways file is empty, because there is no underlying OSM data in that location.

rgreinho:
On way to fix this is to remove the country named "Georgia" from the Geofabrik client since we analyze any city in this country.

However if we succeed to fix #20 and replace the Geofrabrik/BBBike clients with osmnx, this but should go away naturally.

PR #948, opened 2025-09-30
Remove Country Georgia from Geofabrik

rgreinho:
In the GeoFabrik client, the country Georgia and the US Georgia have
identical names, therefore the US state is never found because it
appears later in the list (Europe/Georgia vs NorthAmerica/Georgia).

Since we do not analyze cities in the country Georgia, we can simply
remove it.

For reference this issue was reported upstream in Dec 2021:
https://github.com/pyrosm/pyrosm/issues/162

Fixes PeopleForBikes/brokenspoke-analyzer#938


gstonecipher (approved):
👍

Issue #942, opened 2025-09-24
Apostrophes in City Names


gstonecipher:
When running the wrapper tool for the measure component of the brokenspoke-analyzer, cities with apostrophes in the name result in the following error:

subprocess.CalledProcessError: Command '['psql', '-c', "\\copy residential_speed_limit TO '/Users/gracestonecipher/brokenspoke-analyzer/results/united states/illinois/o'fallon/25.09/residential_speed_limit.csv' WITH (FORMAT CSV, HEADER);", 'postgresql://postgres:postgres@localhost:5432/postgres']' returned non-zero exit status 1.

Additionally, in the folder structure brokenspoke-analyzer/results/<state>/<city>, a file is created for the portion of the name before the apostrophe:

<img width="744" height="82" alt="Image" src="https://github.com/user-attachments/assets/54bbd2eb-03e9-4800-b4e2-21d32d3cbfe8" />

Cities that fail include:

- O'Fallon, IL
- Coeur d'Alene, ID
- Land O'Lakes, FL
- O'Fallon, MO
- Lee's Summit, MO

PR #949, opened 2025-09-30
Escape filenames in SQL export commands

rgreinho:
Ensures special characters are properly escaped in filenames used in SQL
commands.

Fixes PeopleForBikes/brokenspoke-analyzer#942


gstonecipher (approved):
👍

PR #956, opened 2025-10-13
Use Census Bureau boundaries for US cities

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

Based on the comment in https://github.com/PeopleForBikes/brokenspoke-analyzer/issues/886#issuecomment-3321881119, this proposes using Census Bureau place data for boundaries. 

Because the files are downloaded at the state level, caching probably makes sense. The `pygris` package does support it, but does not appear to support specifying the cache directory and will use the platform cache directory. I copied the logic used for caching osmnx.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ RUN apt-get update \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 ENV BNA_OSMNX_CACHE=0
+ENV BNA_PYGRIS_CACHE=0
 WORKDIR /usr/src/app
 COPY --from=builder /usr/src/app/deps ./pkg/deps
 COPY --from=builder /usr/src/app/dist ./pkg/dist
--- a/brokenspoke_analyzer/cli/prepare.py
+++ b/brokenspoke_analyzer/cli/prepare.py
@@ -86,6 +86,7 @@ def prepare_cmd(
             city_speed_limit=city_speed_limit,
             city=city,
             country=country,
+            fips_code=fips_code,
             data_dir=data_dir,
             lodes_year=lodes_year,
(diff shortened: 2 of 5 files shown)


rgreinho:
Thanks a lot for the work @mitchellhenke! ❤️ 

We're still debating internally whether to use only OSM or not, so I'll mark the PR as "do not merge" for now.

Just so you know, we're also working on a way to cache OSM file, piggy backing on what we implemented for the US census files. Our goal is to have caching enabled by default for every file (where it makes sense) for v3, while making sure not to break our cloud pipelines.

mitchellhenke:
Totally understand! I think for my usage, this change has been helpful. With mostly working on a single state, and the file being downloaded per-state, things move slightly faster. The boundaries being guaranteed to exist and not having to worry about getting the wrong boundary has been helpful too.

If there's anything I can address in it, just let me know!

rgreinho:
@mitchellhenke we decided to move forward with the use of `pygris` for fetching the US city boundaries!

So I have 3 minor remarks regarding the PR:

- We will refine the caching mechanism later, but for now it will work fine.
- Can you rename the environment variable `BNA_CENSUS_BOUNDARIES_CACHE` to `BNA_PYGRIS_CACHE`?
- in `analysis.py`, line 144 (`raise TypeError("invalid result class: {gdf_class}")`, there is a bug in the existing codebase. The "f" for the f-string is missing (it should be `f"invalid result class: {gdf_class}"`). If you could fix it at the same time that would be great.

And of course, rebase on top of the v3-base branch.

Thank you very much.

mitchellhenke:
@rgreinho awesome! I've just pushed up the changes to try to address those issues.

Issue #943, opened 2025-09-24
Ypsilanti, MI


gstonecipher:
When running the brokenspoke-analyzer for Ypsilanti, MI, the following error is produced:

subprocess.CalledProcessError: Command '['pgsql2shp', '-u', 'postgres', '-P', 'postgres', '-h', 'localhost', '-p', '5432', '-f', '/Users/gracestonecipher/brokenspoke-analyzer/results/united states/michigan/ypsilanti/25.09/neighborhood_ways.shp', 'postgres', 'neighborhood_ways']' returned non-zero exit status 1.

Investigation of the created files shows that the city boundary (/Users/gracestonecipher/brokenspoke-analyzer/data/ypsilanti-michigan-united-states/ypsilanti-michigan-united-states.shp) looks as expected:

<img width="914" height="703" alt="Image" src="https://github.com/user-attachments/assets/1a26fe33-7a2b-4703-aa00-e54d93dff99e" />

Additionally, the data pulled from OSM (/Users/gracestonecipher/brokenspoke-analyzer/data/ypsilanti-michigan-united-states/ypsilanti-michigan-united-states.clipped.osm) appears to be normal as well:

<img width="741" height="552" alt="Image" src="https://github.com/user-attachments/assets/94411bd2-8332-4563-9720-666f550c6b38" />

It is unclear why the neighborhood_way.shp layer is empty.

rgreinho:
Reproduce with:

```
uv run bna run-with compose --lodes-year=2021 "united states" "ypsilanti" "michigan" 2689140
```

rgreinho:
Command `pgsql2shp', '-u', 'postgres', '-P', 'postgres', '-h', 'localhost', '-p', '5432', '-f',
'/Users/rgreinhofer/projects/PeopleForBikes/brokenspoke-analyzer/results/united states/michigan/ypsilanti/25.09/ neighborhood_ways.shp', 'postgres', 'neighborhood_ways'` failed with:

```
Initializing...
ERROR: Could not determine table metadata (empty table)
```

rgreinho:
From PGCLI, we can indeed see that the table is empy:

```bash
postgres@localhost:postgres> select count(*) from neighborhood_ways;
+-------+
| count |
|-------|
| 0     |
+-------+
SELECT 1
Time: 0.021s
```

rgreinho:
Repro steps:

```bash
[...]

PR #971, opened 2025-11-06
Fix osm2pgrouting 3

### LINKED ISSUE #969: osm2pgRouting 3.0.0 breaks the SQL tables
# Bug report

## Current Behavior

If [osm2pgRouting](https://pgrouting.org/docs/tools/osm2pgrouting.html#) gets updated to v3.0.0, it breaks the `prepare_table.sql` script.

## Possible Solution

As per their [release page](https://github.com/pgRouting/osm2pgrouting/releases/tag/v3.0.0), this is a breaking update, and indeed it breaks our stuff.

We should update the SQL import scripts to switch to `pgrouting`  v3.0.0 and update the documentation for the brokenspoke-analyzer v3, mentioning that is requires `pgrouting` or above.

mitchellhenke:
# Pull-Request

  change)
- Code cleanup / Refactoring
- Documentation

## Description

I was testing a minimal version of these changes for #943 with the dev version of `osm2pgrouting` and now that it's been released, I think this should address the breaking changes. I'm not 100% sure if there are additional columns that should be dropped though.

## Checklist

Fixes #969

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/prepare_tables.sql
+++ b/brokenspoke_analyzer/scripts/sql/prepare_tables.sql
@@ -28,31 +28,27 @@ ALTER TABLE neighborhood_ways DROP COLUMN target_osm;
 ALTER TABLE neighborhood_ways DROP COLUMN priority;
 ALTER TABLE neighborhood_ways DROP COLUMN one_way;
 
-ALTER TABLE neighborhood_ways_intersections DROP COLUMN cnt;
-ALTER TABLE neighborhood_ways_intersections DROP COLUMN chk;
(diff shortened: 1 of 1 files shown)


rgreinho:
Thanks a lot @mitchellhenke for the fix! I'll test it locally as soon as I can. 

However we also have to think about how we're going to update the container and the CI. `osm2pgrouting` just got accepted into [Debian Sid](https://tracker.debian.org/pkg/osm2pgrouting), so it is likely that we will have to switch to sid. But this would also be a great opportunity to upgrade to PosgreSQL 19 and postGIS 3.6.0.

mitchellhenke:
@rgreinho Sounds great, thank you!

PR #977, opened 2025-11-13
Fix osm2pgrouting 3

rgreinho:
Update Dockerfile

Updated the Dockerfile to build `osm2pgrouting` 3.0.0 and use Debian
Trixie as a base image.

Built on top of #971.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -15,8 +15,40 @@ env:
   TEST_FIPS: "3570670"
 
 jobs:
+  osm2pgrouting3:
+    runs-on: ubuntu-latest
+    steps:
+      - name: install dependencies
+        run: |
+          sudo apt-get update
+          DEBIAN_FRONTEND=noninteractive sudo apt-get install -y --no-install-recommends \
+          build-essential \
+          cmake \
+          expat \
+          git \
+          libboost-dev \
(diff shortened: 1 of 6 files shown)


lalver1 (changes requested):
Looks good @rgreinho!

Just requesting for the `dev` stage of the Dockerfile to look like:

```docker
FROM main AS dev
RUN apt-get update && apt-get install -y --no-install-recommends \
  build-essential \
  curl \
  gcc \
  git \
  graphviz \
  just \
  locales \
  npm \
  openssh-client \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
RUN deluser --remove-home bna \
  && delgroup --only-if-empty bna \
  && chown -R root:root /usr/src/app \
  && pip install uv \
  && useradd --create-home --shell /bin/bash bna
RUN curl -o /etc/bash_completion.d/git-completion.bash \
  https://raw.githubusercontent.com/git/git/master/contrib/completion/git-completion.bash
USER bna
```

Since we changed how we install `just` in the `builder` stage.

So in the diff of the Dockerfile you'll see these 2 changes:
deleted `COPY --from=builder /usr/src/app/just /usr/local/bin/`
added `just \`

lalver1 (approved):
Thanks @rgreinho! It's working well now!

Issue #972, opened 2025-11-06
Puerto Rico fails due to missing LODES


gstonecipher:
# Bug report

## Current Behavior

Running any cities in Puerto Rico currently fails due to a lack of LODES data. Since Puerto Rico is part of the US, the analyzer is expecting LODES data to exist, but it doesn't.

## Expected Behavior
Puerto Rico should function like other international cities where employment data is ignored.

## Possible Solution
In the old BNA, when running international cities, there is a checkbox to say "omit jobs data in analysis" that must be checked for any cities in Puerto Rico to run - can we add some option here? I believe this would already be the default for other international cities.

## Steps to Reproduce
brokenspoke-analyzer % uv run bna run-with compose --with-parts=measure "united states" "san juan" "puerto rico" 7276770

PR #978, opened 2025-11-14
Implement an exception for Puerto Rico

rgreinho:
Even though Puerto Rico is part of the US, the US Census Bureau does not
collect any LODES data for it. As a result an exception was created to
prevent the analyzer to attempt to fetch the LODES data.

Fixes PeopleForBikes/brokenspoke-analyzer#972


lalver1 (approved):
👍

PR #1007, opened 2026-01-19
Ignore width for footpaths where bike=designated

gstonecipher:
When bicycle=designated, bikes are explicitly supposed to ride on these paths, so the width should not matter

# Pull-Request

## Types of changes

## Description
This fix ignores the width of footpaths where bicycle=designated. If it's designated, it should mean that bikes are explicitly supposed to ride there (Shared Use Path), or in non-US countries, it could mean specifically signposted bike/ped paths. Consider https://community.openstreetmap.org/t/use-of-bicycle-designated-vs-bicycle-yes-outside-of-germany/3230

Testing:
McFarland, WI

uv run bna run-with compose --with-parts=features usa mcfarland wisconsin 5546850

￼
Before fix: Footpath is bicycle=designated by width = 1.5m does not get classified as a path
Pasted Graphic

￼
After fix: Footpath gets classified as a path
Pasted Graphic 1

## Checklist


lalver1 (approved):
The footway logic looks good to me! I noticed that the `uv.lock` file changed but I guess that is ok.

PR #1018, opened 2026-02-02
Calculate overall score by weighted census blocks

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

This is a rather significant proposal, but I think worth considering given the changes since the last score release. Currently, within each score category the score is weighted by population, but the overall scores are weighted by each category score, not each census block overall score.

This can lead to some more surprising distributions of scores. In the extreme example, if there were a city with a single transit destination in which 1% of the population were within biking distance, but only 5% of that 1% could safely bike to it, the score would be 5/100 for transit. Transit is weighted at 15% of the overall score, which means the city would have .75/15 for the transit portion of the score. This results in a score that is more heavily weighted towards transit despite only 5% of the population being within biking distance of transits.

As a case study of a real city, Milwaukee's transit score is 3/100, however there are only 3 transit stations with around 20% of the city having access to at least one of them. This results in .45 out of a possible 15 for the transit portion of the score in Milwaukee.

My above examples are limited to low score cases, but the reverse is also possible.

My proposal here is to change overall scores so that they're weighted by each census block's overall score and population which I think can result in a more representative overall score. The individual category scores would still score access to those destinations regardless of how many people in the city are within reach of them.

## Checklist

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -117,7 +117,7 @@ jobs:
 
           echo "Ensure neighborhood_overall_scores table contains 23 records."
           ACTUAL=$(psql -t -c "SELECT COUNT(*) FROM neighborhood_overall_scores;" $DATABASE_URL)
-          test $ACTUAL -eq 23
+          test $ACTUAL -eq 24
 
   docker-integration:
     runs-on: ubuntu-latest


rgreinho:
> I added a new score instead of replacing the existing one to make it easier to test, but it's resulting in a test failure.

Oh it is simply because we expect 23 columns in the output, but as you added one there are now 24.

The change would be to replace this line: https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/.github/workflows/e2e.yaml#L120 to `test $ACTUAL -eq 23`.

PR #1051, opened 2026-03-20
Remove routable unpaved ways

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description
(This is a re-do of #1042)

Currently all ways are included, regardless of surface type. In limited investigation, most often these are paths, though the potential to include unpaved roads is also present.

I wasn't sure on the desired impact on `neighborhood_paths` destinations, so the proposed changes here only deletes `neighborhood_ways` used to route. This means that unpaved paths will still be included as potential path/trail destinations.

In choosing which values to exclude, I used [taginfo](https://taginfo.openstreetmap.org/keys/surface#values) to choose the most common values that are not paved.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/core/compute.py
+++ b/brokenspoke_analyzer/core/compute.py
@@ -73,6 +73,32 @@ def features(
         sql_script = sql_feature_script_dir / script
         dbcore.execute_sql_file(engine, sql_script)
     sql_script = sql_feature_script_dir / "paths.sql"
+
+    logger.info("Remove routable unpaved ways")
+    dbcore.execute_query(
+        engine,
+        """
+        DELETE FROM neighborhood_ways w
+        USING neighborhood_osm_full_line o
+        WHERE
+            w.osm_id = o.osm_id AND surface IN (
+                'dirt',
+                'unpaved',
+                'sand',
+                'compacted',
+                'ground',
+                'rock',
+                'soil',
+                'earth',
+                'gravel',
+                'woodchips',
+                'fine_gravel',
+                'mud',
+                'grass'
+            );
+        """,
+    )
+
     bind_params = {"nb_output_srid": output_srid}
     execute_sqlfile_with_substitutions(engine, sql_script, bind_params)
     sql_scripts = [


gstonecipher:
Hi Mitchell, thanks for this PR. 

Methodology wise, I'm torn on this one. Here in Bend, we have quite a few unpaved trails along the canals that are intended as part of the bike network - though I understand why excluding them could also make sense, given that all riders may not be comfortable on other surfaces. In OSM, these are tagged specifically as highway=cycleway, though I'm seeing some dirt trails here that are also tagged as cycleways, which are definitely not accessible.

I'm going to do some more investigation of specific surface types to see if we can better separate out different types of paths, especially MTB trails.

Do you have other examples that you were looking at that should be excluded from routing?

I'm also going to fully separate the neighborhood paths destinations from the bicycle infrastructure paths, because the current setup means that we miss a ton of trails when highway=footpath.

mitchellhenke:
Yeah, it's not super clear cut. The easy line to draw is excluding all unpaved, but like you mentioned, some unpaved paths may be a true part of a city's network. Some values are more plausibly "official" and usable like `compacted`, but ones like `sand`, `mud`, `woodchips`, `grass`, etc. are much less likely to be. The [informal](https://wiki.openstreetmap.org/wiki/Tag:informal%3Dyes) tag may be useful here as well.

As an example from Milwaukee, there is an unpaved informal desire path along the east side of Lincoln Memorial Drive going under the Mason Street bridge ([way 393084273](https://www.openstreetmap.org/way/393084273)). It is included in BNA's analysis as a low-stress route, but it is not a pleasant ride and should probably be excluded. The OSM tags I think make it clear as it is both `surface=mud` and `informal=yes`.

image

gstonecipher:
Excluding that type of path totally makes sense. I'll spend some more time looking into tags and come up with a path forward for next year's analysis.

naterkane:
@mitchellhenke @gstonecipher @lalver1 @rgreinho , 

+1 for this, in some areas it can be quite a big flaw in how an area is calculated. It may be reasonable to tune this a bit further, but I'm wondering where the best place for discussion may be other than this PR's comment thread.

This area scores much higher than it should, thanks to the technical trail in a county park, as well as myriad trails included from Patapsco State Park. Many of those trails are mapped thanks to my ground-truth auditing, GPS, and OSM updates.
SCR-20260924-jpjr

I don't believe this is just a matter of unpaved vs paved, however; due to this issue is general, it makes the scoring entirely unreliable in my part of Central Maryland and looks like it will prevent adoption as a metric we can target for infrastructure improvements in my county. 

Many of the trails considered low stress in the above example should be excluded, however if a path _is_ tagged `mtb:scale=0`[link](https://wiki.openstreetmap.org/wiki/Key:mtb:scale#mtb:scale-0), `mtb:scale=0-`, or `mtb:scale:imba=0`[link](https://wiki.openstreetmap.org/w/index.php?title=Tag:mtb:scale:imba%3D0&action=edit&redlink=1) it may be worth considering it fair to be included without regard to the value of the `surface` tag, since those tags imply a reasonably compacted and rideable surface. 

Taking language and regional variations into consideration, what might be the best way to draft a ruleset that captures how this update should be applied?

PR #1057, opened 2026-03-20
fix two-way cycle tracks

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

(This is a re-do of #1012 as it was merged into v3-base and didn't make it into main)

There is a bug in the changes from https://github.com/PeopleForBikes/brokenspoke-analyzer/pull/890 that results in two-way cycle tracks in being classified as high stress in the "both ways" cases.

An example is way [1199660627](https://www.openstreetmap.org/way/1199660627#map=20/43.0507475/-87.9033802&layers=C), which is currently designated as high stress but has a two-way cycle track. Tagging that implies two-way `cycleway:left/right:oneway` with an associated `cycleway:left/right=track` should always result in a low-stress designation, but the query as-is does not catch them.

To fix it, I'm proposing setting the infra to `track` if one side is tagged as track, and there are tags indicating the cycleway is two-way.

## Checklist

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
+++ b/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
@@ -32,20 +32,18 @@ SET
         WHEN
             (
                 osm."cycleway:right" = 'track'
-                AND osm."oneway:bicycle" = 'no'
                 AND (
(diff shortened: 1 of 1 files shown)


lalver1 (approved):
Thank you!

PR #1059, opened 2026-03-25
Expand the exporter to support more object stores

rgreinho:
Leverages the Obstore library in order to be able to export results to
more data stores.

The S3 export was rewritten and the R2 export was added.

### DIFF
--- a/brokenspoke_analyzer/cli/export.py
+++ b/brokenspoke_analyzer/cli/export.py
@@ -1,18 +1,63 @@
 """Define the export sub-command."""
 
+import asyncio
 import pathlib
-import typing
 
 import rich
 import typer
+from click.core import ParameterSource
 from loguru import logger
+from obstore import store
 from typing_extensions import Annotated
 
 from brokenspoke_analyzer.cli import common
 from brokenspoke_analyzer.core import exporter
 
+# import typing
+
+
 app = typer.Typer()
 console = rich.get_console()
+
+
+def allow_only_env_var(
+    ctx: typer.Context, param: typer.CallbackParam, value: str
+) -> str:
+    """Allow only environment variables for a parameter."""
+    if param.name is None:
+        raise ValueError(f"{param.name} must be set.")
+    if ctx.get_parameter_source(param.name) != ParameterSource.ENVIRONMENT:
+        raise typer.BadParameter("only allowed as environment variable")
+    return value
+
+
+CloudflareAccountID = Annotated[
+    str,
+    typer.Option(
+        help="Cloudflare Account ID",
+        envvar="CLOUDFLARE_ACCOUNT_ID",
+        callback=allow_only_env_var,
+    ),
+]
+R2AccessKeyID = Annotated[
+    str,
+    typer.Option(
+        help="R2 Access Key ID",
+        envvar="R2_ACCESS_KEY_ID",
+        callback=allow_only_env_var,
+    ),
+]
+R2SecretAccessKey = Annotated[
+    str,
+    typer.Option(
+        help="R2 Secret Access Key",
+        envvar="R2_SECRET_ACCESS_KEY",
+        callback=allow_only_env_var,
+    ),
+]
+WithBundle = Annotated[
+    bool, typer.Argument(help="bundle all the files in a zip archive")
+]
 
 
 @app.command()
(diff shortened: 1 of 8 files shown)


rgreinho:
Ready for review @lalver1 🙂

brokenspoke_analyzer/cli/export.py
```diff
@@ -1,11 +1,14 @@
 """Define the export sub-command."""
 
+import asyncio
 import pathlib
 import typing
```
lalver1: Little cleanup since we are not using it anymore in this file

```suggestion

```

brokenspoke_analyzer/cli/export.py
```diff
@@ -54,25 +97,67 @@ def s3(
...
-        folder = exporter.s3_directories(bucket_name, s3_dir)
-        exporter.s3(database_url, bucket_name, folder, with_bundle)
-        return folder
+        return asyncio.run(s3_custom_(database_url, bucket_name, s3_dir, with_bundle))
+
+
+@app.command()
+def r2(
+    database_url: common.DatabaseURL,
+    account_id: CloudflareAccountID,
```
lalver1, line 122: I think these parameters (`account_id, `access_key_id`, and `secret_access_key`) are not needed anymore since authentication credentials are stored in env variables?

brokenspoke_analyzer/cli/export.py
```diff
@@ -54,25 +97,67 @@ def s3(
...
+    - R2_SECRET_ACCESS_KEY
+    """
+    with console.status("[green]Uploading results to Cloudflare R2..."):
+        asyncio.run(r2_(database_url, bucket_name, country, city, region, with_bundle))
+
+
+@app.command()
+def r2_custom(
+    database_url: common.DatabaseURL,
+    account_id: CloudflareAccountID,
```
lalver1, line 146: Similar situation as above, these parameters probably aren't needed anymore?

brokenspoke_analyzer/cli/run_with.py
lalver1: Looks like pandas is not used anymore
```suggestion

```

lalver1, line 12: Not used anymore, looks like we use `rich.get_console()` now
```suggestion

```

lalver1: Looks like we don't use `importer` here anymore
```suggestion

```

lalver1: Same with `runner`
```suggestion

```

brokenspoke_analyzer/cli/run.py
lalver1: Same, looks like we don't use `dbcore` here anymore
```suggestion

```

brokenspoke_analyzer/core/exporter.py
```diff
@@ -1,19 +1,49 @@
...
 import pathlib
 import shutil
 import tempfile
 import typing
 from datetime import date
 from enum import Enum
+from typing import TYPE_CHECKING
 
 import boto3
+import boto3.s3
```
lalver1: Little cleanup
```suggestion

```

brokenspoke_analyzer/core/exporter.py
```diff
@@ -1,19 +1,49 @@
...
+import obstore
```
lalver1: Same here
```suggestion

```

lalver1 (changes requested):
Nice! This looks good to me @rgreinho! I wasn't able to actually test the new commands but the logic looks good. I only noticed a few cleanups that are not critical, just thought to point them out though

rgreinho:
Oh wow! That's a lot of cleanup! Thanks a lot @lalver1, I'll address them all, but I am wondering why ruff/mypy did not catch them automatically... 🤔

brokenspoke_analyzer/cli/export.py
rgreinho: Addressed. ✅

brokenspoke_analyzer/cli/export.py
rgreinho: Ah these are a bit special. They are required, but we only use their values from the environment variables, because that is what the underlying framework (`obstore`) does.

So when defining them I attached a callback:

```py
CloudflareAccountID = Annotated[
    str,
    typer.Option(
        help="Cloudflare Account ID",
        envvar="CLOUDFLARE_ACCOUNT_ID",
        callback=allow_only_env_var,
    ),
]
```

Which checks that these parameters are set, but via environment variables only:

```py
def allow_only_env_var(
    ctx: typer.Context, param: typer.CallbackParam, value: str
) -> str:
    """Allow only environment variables for a parameter."""
    if param.name is None:
        raise ValueError(f"{param.name} must be set.")
    if ctx.get_parameter_source(param.name) != ParameterSource.ENVIRONMENT:
        raise typer.BadParameter("only allowed as environment variable")
    return value 
```

So the parameters themselves are not used directly, so that we can avoid wiring them down to the functions creating the stores, but it guarantees that the values have been provided via the right environment variables. If not the function will fail with an appropriate error message.

That's a bit of a trick, but I though it would be a nice guardrail. It is not perfect because it only applies to the `export` commands, and not to the `run` or `run-with` ones. But that should be a good start already.

If you find it confusing, I can rethink it though, but I would really like to have a mechanism ensuring that we have the right parameters

brokenspoke_analyzer/cli/export.py
rgreinho: Same as bellow, so I'll detail why there.

brokenspoke_analyzer/cli/run_with.py
rgreinho: Addressed. ✅

rgreinho: Addressed. ✅

rgreinho: Addressed. ✅

brokenspoke_analyzer/cli/run.py
rgreinho: Addressed. ✅

brokenspoke_analyzer/core/exporter.py
rgreinho: Addressed. ✅

brokenspoke_analyzer/core/exporter.py
rgreinho: Addressed. ✅

rgreinho:
> Oh wow! That's a lot of cleanup! Thanks a lot @lalver1, I'll address them all, but I am wondering why ruff/mypy did not catch them automatically... 🤔

Arg, it is because we misconfigured ruff...

In `pyproject.toml`, the 

```toml
[tool.ruff.lint]
select = ["D"]
```

Should have been:

```toml
[tool.ruff.lint]
select = ["ALL"]
```

To be in strict mode.

See <https://docs.astral.sh/ruff/linter/#rule-selection> for reference. 

Once we merge #1059, #1069, #1070,  and #1075, I will create a PR to set `ruff` in strict mode. Maybe in a progressive way if there are too many fixes required.

lalver1 (approved):
Looks great @rgreinho!

Thanks for explaining why the parameters `account_id`, `access_key_id`, and `secret_access_key` are needed, I agree that it is best to ensure that values are provided for them 👍 

Oh I see! Yeah we can make a followup later to run ruff and catch any other cleanups, thanks!

PR #1063, opened 2026-04-03
Honor --with-bundle flag for all exports

rgreinho:
The `--with-bundle` flag was ignored for non-local exports. This patch
ensures it now works for all types of exports.

Drive-by:
- Fixes the optional argument syntax in the Typer commands.


gstonecipher (approved):
👍

PR #1075, opened 2026-04-14
Fix Dependabot Configuration

rgreinho:
Fixes the configuration file for Dependabot to use the appropriate
`package-ecosystem` value as well as a `cooldown` value to prevent
ultra-new packages to be added right away.


lalver1 (approved):
👍

PR #1076, opened 2026-04-20
Configure ruff in very strict mode

rgreinho:
Configures `ruff` to perform very strict checks.

Some rules where disabled as they either did not make sense at all for
us or may be addressed in the future.


lalver1 (approved):
Nice! Thanks @rgreinho!

PR #1078, opened 2026-04-25
Add WorldPop SourceAdapter

### LINKED ISSUE #1061: Support using census data for cities in Canada
# Feature request

Support using census data, instead of synthetic data, for analyzing Canadian cities.

## Current Behavior

The analyzer uses [synthetic data generated by this method](https://peopleforbikes.github.io/brokenspoke-analyzer/workflow.html#extra-steps-for-non-us-cities) for analyzing cities in Canada.

## Expected Behavior

The analyzer should use census data instead of synthetic data for more accurate results.

## Possible Solution

Two possible data sources are [Statistics Canada](https://www.statcan.gc.ca/en/start) and [WorldPop](https://hub.worldpop.org/project/categories?id=3). A high-level design for this feature using Statistics Canada data could be:

- In the PFB public documents S3, create a bucket structure (similar to how speed limits are currently stored) for
  - [Boundary files (Census metropolitan areas and census agglomerations)](https://www12.statcan.gc.ca/census-recensement/alternative_alternatif.cfm?l=eng&dispext=zip&teng=lcma000b21a_e.zip&k=%20%20%20%2013083&loc=//www12.statcan.gc.ca/census-recensement/2021/geo/sip-pis/boundary-limites/files-fichiers/lcma000b21a_e.zip)
  - [Census population (dissemination blocks)](https://www12.statcan.gc.ca/census-recensement/alternative_alternatif.cfm?l=eng&dispext=zip&teng=ldb_000b21a_e.zip&k=%20%20%20309953&loc=//www12.statcan.gc.ca/census-recensement/2021/geo/sip-pis/boundary-limites/files-fichiers/ldb_000b21a_e.zip)
- This S3 bucket will serve as the source of truth for boundary and population data for cities in Canada. This approach is necessary because census data provided by Statistics Canada is only available for the whole country, these data are not available on a province by province basis, this requiring pre-processing to prepare the data set.
- For convenience, provide a list of the cities in Canada whose data is available in S3.
- In [`core/analysis.retrieve_city_boundaries`](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/brokenspoke_analyzer/core/analysis.py#L147) add another conditional to use the data source in S3 described above if the city being analyzed is in Canada.
- The previous step can be refactored into [core/datastore.BNADataStore](https://github.com/PeopleForBikes/brokenspoke-analyzer/blob/main/brokenspoke_analyzer/core/datastore.py#L65).

lalver1:
Closes #1061

This PR adds a [WorldPop](https://hub.worldpop.org/project/categories?id=3) `SourceAdapter` to work with WorldPop data. The immediate need for this is to run analyses for Canadian cities using the [2021 WorldPop 1km resolution population data](https://hub.worldpop.org/geodata/summary?id=76683).

## Highlights

- The `WorldPopAdapter` works for countries for which the first 3 characters of the long name of the country matches its [ISO 3166 country code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes). This is the case for Canada (CAN), but it may not be true for other countries. To make the `WorldPopAdapter` truly generic, we'd need to add a utility function to convert a country's long name (as used by Nominatim) to its ISO 3166 country code. This would certainly make the code DRYer, but maybe we come back to this refactor after June?
- The `year` parameter of the `WorldPopAdapter`'s `__init__` method can be used to select another year's WorldPop dataset.
- New packages 
  - The [`rasterio`](https://github.com/rasterio/rasterio) package was added to load the geoTIFF file. I guess it'd be nice to find another package with a more straightforward license, but `rasterio` seems to be the standard for Python projects.
  - [`numpy`](https://numpy.org/) was added to make the generation of the `GEOID20` column faster since the Shapefile for the population data may turn out to have a bunch of rows.
  
## Reviewing

I used `uv run bna -vvv run "Canada" "moncton" "new brunswick"` to test out a small city.


brokenspoke_analyzer/core/datasource.py
```diff
@@ -165,6 +171,118 @@ def validate(self, datastore: pathlib.Path) -> None:
...
+    def urls(self) -> abc.Sequence[yarl.URL]:
+        """Return the source data URLs."""
+        return [
+            yarl.URL(
+                self.source_url
+                / self.year
+                / self.country
+                / "v1/1km_ua/constrained"
+                / str(f)
+            )
```
rgreinho: Since `self.SOURCE_URL` is already a `yarl.URL` object, there is no need to reuse the constructor. The `/` operator will automatically append the parts and return a `yarl.URL` object.

```suggestion
                self.source_url
                / self.year
                / self.country
                / "v1/1km_ua/constrained"
                / str(f)
```

brokenspoke_analyzer/core/datastore.py
```diff
@@ -241,6 +241,18 @@ async def download_2020_census_blocks(
         s = datasource.CensusAdapter(fips, self.mirror)
         await self.fetch_from_source(session, s, cache_only=cache_only)
 
+    async def download_worldpop(
+        self,
+        session: aiohttp.ClientSession,
+        country: str,
+        year: str = "2021",
```
rgreinho, line 248: Can we document why we default to 2021?

brokenspoke_analyzer/core/datasource.py
```diff
@@ -165,6 +171,118 @@ def validate(self, datastore: pathlib.Path) -> None:
...
+        """Prepare the data files."""
+        if len(self.files) != 1:
+            raise ValueError(
+                f"only 1 file was expected, {len(self.files)} found: {self.files}"
+            )
+        file_geotiff = datastore / self.files[0]
+        output_dir = datastore.resolve()
+        file_shp = output_dir / "population.shp"
+
+        if not file_shp.is_file():
```
rgreinho: Nitpicking a little, but I would have reverted this condition to return early and avoid intending mostly the whole function:

```py
if file_shp.is_file():
   return

# Otherwise Create the shapefile.
````

**Review by rgreinho — CHANGES_REQUESTED** [review]
Love it @lalver1! 🤗 

Just a couple minor remarks and a question, but aside from that it is good to merge 👍

File: brokenspoke_analyzer/core/datasource.py (hunk 3 above)
Comment (rgreinho) [reply brokenspoke_analyzer/core/datasource.py]: Or even:
```py
if file_shp.exists():
```

brokenspoke_analyzer/core/datastore.py
lalver1: Yep! Good point, I'll add that it corresponds to the latest official 2021 Canadian Census.

brokenspoke_analyzer/core/datasource.py
lalver1: Not at all, good suggestion, I'll re-do the condition 👍

lalver1:
Thanks for the comments @rgreinho! This should be ready for a re-review but let me know if there's anything else!

rgreinho (approved):
All good! 👍

PR #1107, opened 2026-06-05
Fix bike infra direction handling for opposite_track and contraflow/bidirectional cycleways on two-way roads

### LINKED ISSUE #1106: Unreachable opposite_track conditions in one_way_car = 'ft' branch
# Bug report

The one_way_car = 'ft' logic in features/bike_infra.sql contains two opposite_track conditions that check for one_way_car = 'tf'. Because these conditions are nested inside the one_way_car = 'ft' branch, they can never evaluate to true. As a result, some opposite_track bicycle infrastructure may not be classified as intended.

## Current Behavior

In features/bike_infra.sql, the one_way_car = 'ft' branch contains the following conditions:

WHEN
    (
        one_way_car = 'tf'
        AND osm."cycleway:left" = 'opposite_track'
    )
    THEN 'track'
WHEN
    (
        one_way_car = 'tf'
        AND osm."cycleway:right" = 'opposite_track'
    )
    THEN 'track'

These conditions are nested inside a CASE branch that is only entered when one_way_car = 'ft'.

As a result, both conditions are unreachable because they require one_way_car to simultaneously be 'ft' and 'tf'. The corresponding THEN 'track' branches can never be executed.

## Expected Behavior

opposite_track infrastructure should be evaluated correctly within the one_way_car = 'ft' branch, and the relevant conditions should be reachable.

## Possible Solution

Remove the one_way_car = 'tf' checks from these conditions so that opposite_track is evaluated based solely on the cycleway tags within the one_way_car = 'ft' branch.

mitchellhenke:
# Pull-Request

## Types of changes

  change)
- Code cleanup / Refactoring
- Documentation

## Description

This fixes the issue identified in #1106, though I ended up finding a couple other issues as well. They are:

- One-way "contraflow" bike lanes on a two-way roads are currently credited going in the wrong direction in BNA analysis. In the US, this less common case happens if there is a bike lane on the leftmost portion of the road, but there is two-way vehicle traffic to the right. An example is [way 619457505](https://www.openstreetmap.org/way/619457505#map=19/41.943516/-87.644927) in Chicago, IL. A screenshot from Google Streetview is below.
- Two-way non-track cycle lanes are not currently handled correctly. Two-way cycle tracks were addressed in #1057, but painted lanes and buffered lanes were not. This is also less common infrastructure, but an example is [way 6027184](https://www.openstreetmap.org/way/6027184#map=19/45.029388/-93.286988) in Minneapolis, MN.

image

image

## Checklist

Fixes: #1106

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
+++ b/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
@@ -216,10 +216,25 @@ SET
                     THEN 'sharrow'
                 WHEN osm."cycleway:right" = 'shared_lane'
                     THEN 'sharrow'
+                -- left-side carrying ft (bidirectional or contraflow on two-way road)
+                WHEN
+                    osm."cycleway:left" = 'shared_lane'
+                    AND osm."cycleway:left:oneway" IN ('yes', 'no')
+                    THEN 'sharrow'
+
                 WHEN osm.cycleway = 'buffered_lane'
                     THEN 'buffered_lane'
-                WHEN osm."cycleway:right" = 'buffered_lane'
+                WHEN
+                    osm."cycleway:right" = 'buffered_lane'
+                    AND COALESCE(
+                        osm."cycleway:right:oneway", 'yes'
+                    ) != '-1'
                     THEN 'buffered_lane'
+                WHEN
+                    osm."cycleway:left" = 'buffered_lane'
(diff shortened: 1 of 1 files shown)


rgreinho:
@mitchellhenke This indeed fixes a bug, but since we are still running the computation for non-US, we must wait until we work on v4 to merge this patch as it would be a breaking change.

PR #1155, opened 2026-07-18
Use Datastore To Fetch City Boundaries

### LINKED ISSUE #1141: Add a new datasource type to fetch and cache city boundaries
### Summary

Add a new datasource type that can fetch and cache city boundary data for use in analysis workflows.

### Problem or need

The project currently lacks a dedicated datasource for city boundaries. In our effort to standardize data sources and data caching we should add a new `datasource` to handle these items.

### Proposed solution

Following the pattern from existing `datasources`, add a new one for city boundaries.

Reuse the logic from the `retrieve_city_boundaries()` function in `brokenspoke_analyzer/core/analysis.py`.

This may require patching the Pygris library though as it comes with its own caching mechanism and prints messages on the screen. Evaluate this aspect before starting to work on the implementation itself. The Pygris library has not received any update since December 2025, so we may have to fork or vendor it if they do not include our changes.

### Alternatives considered

_No response_

### Expected impact

_No response_

rgreinho:
Updates the logic to leverage the Datastore to fetch the city boundaries
file and cache them.

Fixes PeopleForBikes/brokenspoke-analyzer#1141

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ pip-log.txt
 profile.svg
 results/
 results/*
+reviews/
 sdist/
 target/
 var/
--- a/brokenspoke_analyzer/cli/prepare.py
+++ b/brokenspoke_analyzer/cli/prepare.py
@@ -1,7 +1,6 @@
 """Define the prepare sub-command."""
 
 import asyncio
-import logging
 import pathlib
 
 import aiohttp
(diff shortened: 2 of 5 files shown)


rgreinho:
@lalver1 rebased on your worldpop PR, so this should be good to go 👍

rgreinho:
Merging it to be able to move forward with the remaining PRs for the 3.2.0 release.

PR #1163, opened 2026-07-19
Fix low order segment stress

### LINKED ISSUE #1153: Unclassified and residential road stress output doesn't match the methodology
### Version

main

### Current behavior

Residential and unclassified roads are reclassified to tertiary if they have any dedicated bike infrastructure or are more than one lane or have a speed limit greater than 30 mph. This occurs in `brokenspoke_analyzer/scripts/sql/features/class_adjustments.sql`

Other segments not reclassified in the pass above are handled by either
`brokenspoke_analyzer/scripts/sql/stress/stress_segments_lower_order.sql` or `brokenspoke_analyzer/scripts/sql/stress/stress_segments_lower_order_res.sql`
These set the stress based only on the speed limit.

### Expected behavior

According to the [methodology page](https://cityratings.peopleforbikes.org/about/methodology) Table 5b. Segment Stress, Residential and Unclassified Streets, the segment stress relies on all of the following inputs:

- Type of bike infrastructure (if any)
- Speed limit
- Number of lanes (for shared lane conditions)

The methodology doesn't match with how the code currently handles segments. The code currently removes all unclassified and residential segments from the lower order side if any dedicated bike infrastructure exists and nothing in the code for the lower order streets takes into account the number of lanes when there is no dedicated bike infrastructure.

### Steps to reproduce

1. Run a city with a residential or unclassified street that has dedicated bike infrastructure and/or one that has more than one lane in OSM.
2. Note the discrepancy between what the methodology says the stress should be and what the output is.

An example is [way: 417695405](https://www.openstreetmap.org/way/417695405) and [way: 104380852](https://www.openstreetmap.org/way/104380852) which are two adjoining segments of Sports Center Drive in Lexington, KY which is best classified as residential or unclassified as it is a local low order street on UKs campus. [way: 417695405](https://www.openstreetmap.org/way/417695405) has a bike lane and is therefore reclassified as a tertiary road and because of that it is marked as high stress due to the speed limit of 25 mph. The southern portion [way: 104380852](https://www.openstreetmap.org/way/104380852) which basically has the same levels of traffic but is a shared lane is not reclassified and therefore runs through the low order route and with a speed limit of 25 mph is listed as low stress.

### Environment

- OS: MacOS 15

jcedmiston:
# Pull-Request

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
      change)
- [x] Code cleanup / Refactoring
- [ ] Documentation

## Description

This pull requests removes `class_adjustments.sql` and replaces the `stress_segements_lower_order.sql` logic with an inlinable sql function which calculates the stress based on the segments bike infrastructure, speed limit and number of lanes. Using the SQL function prevents code duplication between the `tf` and `ft` cases.

It also simplifies the code needed for residential and unclassified segments by removing unused inputs and by 'coalescing' the `city_default_speed` and `state_default_speed` values in the calling python code, so the sql can then be the same between the two.

This change fixes the stress output for residential and unclassified streets to be inline with what the methodology calls for as well as simplifying the code required to compute the stress. Previously the `class_adjustments.sql` made changes to the segment before the stress calculation making it difficult to determine why a segments stress was computed the way it was when comparing the OSM data to the methodology.

I ran the `just test` command which passed fully. I also ran Santa Rosa, NM and Lexington, KY both of which completed successfully. The later of which was to confirm that the example I provided in PeopleForBikes/brokenspoke-analyzer#1153 matched up with the methodology and this does fix the segments stress. If I need to run anything else, I'd be happy to.

## Checklist

Fixes: PeopleForBikes/brokenspoke-analyzer#1153


rgreinho:
Thanks a lot  @jcedmiston for this patch! 

We are definitely going to review it carefully, but we must make sure that it is not a breaking change as we are still working on publishing the scores for non-US cities. We cannot alter the methodology mid-computation, even if there was a bug in it.

So this patch may end up being merged in v4.0.0, that we aim to release around the end of November 2026.

As a very minor edit, I would suggest using `CREATE OR REPLACE FUNCTION` when creating a new SQL function.

jcedmiston:
@rgreinho No problem, happy to help! Waiting makes total sense, I realized I checked non-breaking fix in the list, but I agree this would cause a change in outcomes. I updated the comment to mark the correct breaking change checkbox.

On a different note, I was thinking about it the other day and realized that the methodology doesn't explicitly state every case for the low class shared lane situation. My proposed changes assume the implicit cases are high stress but that might be something to discuss before this is merged and maybe worthwhile to update the methodology table to show those cases and whatever stress they should be marked as.

| Bike Facility | Speed | Number of Lanes (Each Direction) |              Stress             |
|:-------------:|:-----:|:--------------------------------:|:-------------------------------:|
|      ...      |  ...  |                ...               |               ...               |
|  Shared Lane  |  <=25 |                 1                |               Low               |
|  Shared Lane  |  <=25 |                >1                | No Stress Listed (Assumed High) |
|  Shared Lane  |  >25  |                 1                | No Stress Listed (Assumed High) |
|  Shared Lane  |  >25  |                >1                |               High              |

It would make sense to me if the low order share lane case followed the same logic as the high order stress but with a speed cut off of 25 instead of 15, like this:
| Bike Facility | Speed | Number of Lanes (Each Direction) | Stress |
|:-------------:|:-----:|:--------------------------------:|:------:|
|      ...      |  ...  |                ...               |   ...  |
|  Shared Lane  |  <=25 |                 1                |   Low  |
|  Shared Lane  |  <=25 |                >1                |  High  |
|  Shared Lane  |  >25  |                 -                |  High  |

PR #1164, opened 2026-07-21
Add Agent Files

rgreinho:
Adds files for persistent context used by agents.


rgreinho:
This will file is the initial one, generated by `/claude init`, with some minor adjustments from me after reviewing it.

Merging it as I am planning to experiment some workflows with Claude code.

PR #1178, opened 2026-07-28
Add Place and County Sudivision Data Sources

### LINKED ISSUE #1177: PyGRIS cache behavior broken
### Version

main

### Current behavior

When fetching the city boundaries for US cities, the `pygris` library fetches the US census files per state and caches them. Then it will process them to extract the specific boundaries for a city.

PR #1155 introduces a regression where only the final product is being cached, resulting in every FIPS-code lookup to re-download the entire state's Census TIGER shapefile with no reuse.

### Expected behavior

We should preserve the caching behavior of the US census files.

One way to do so, is to create new datatsource adapters for the US census place and the US census county sub-division files.

### Steps to reproduce

N/A

### Environment

- Python 3.14.6
- 8b90a1b Add WorldPop support for all/supported countries (#1167)

rgreinho:
Adds 2 new source adapters to cache the US census file for the TIGER
places and county sub-division files.

This will eventually replace the use of the PyGRIS library.

Fixes PeopleForBikes/brokenspoke-analyzer#1177

### DIFF
--- a/brokenspoke_analyzer/core/datasource.py
+++ b/brokenspoke_analyzer/core/datasource.py
@@ -23,6 +23,9 @@ from brokenspoke_analyzer.core import (
 )
 from brokenspoke_analyzer.core.utils import unzip
 
+# TIGER base URL -- Topologically Integrated Geographic Encoding and Referencing.
+TIGER_URL = yarl.URL("https://www2.census.gov/geo/tiger/")
+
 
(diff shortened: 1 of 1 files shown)


rgreinho:
This PR depends on PR #1155. Once #1155 is merged, I'll update the `datastore.download_city_boundaries()` function to use these new data sources.

PR #1193, opened 2026-08-09
Add improved protected bike lane identification

### LINKED ISSUE #1194: Add in Protected Bike Lane (PBL) auto-detection
### Problem or need

Since OpenStreetMap (OSM) tags can be ambiguous, PBLs are sometimes incorrectly categorized as "Off-Street Paths" in the analyzer. The reason is that a PBL, in OSM terms, is a "track" (separate from the road, i.e., off-road; typically separated by curbs, parking lots, grass, etc.). A "track" can be modeled in OSM as:

1- Draw a separate way along the roadway and tag the separate way as[ highway](https://wiki.openstreetmap.org/wiki/Key:highway)=[cycleway](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcycleway)
2- Add a[ cycleway](https://wiki.openstreetmap.org/wiki/Key:cycleway)=[track](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dtrack) tag to the existing way that represents the roadway
 
What may end up happening is that a PBL will be categorized as an "Off-Street Path" in the analyzer when it has been modeled using method 1 above (the "ft_bike_infra" or "tf_bike_infra" fields are "NULL" and so they get classified according to `WHEN "XWALK" IS NULL AND "FUNCTIONAL" = 'path' AND "FT_BIKE_IN" IS NULL THEN 'Off-Street Path'`).

The brokenspoke-analyzer should be able to identify off-street paths that should more accurately be classified as PBLs instead.

These are two examples of PBLs incorrectly identified as "Off-Street Paths" in the analyzer:

- Wauwatosa, WI (North 87th St.)
  Image

- Austin, TX (West 3rd St.)
  Image

### Proposed solution

@gstonecipher sketched out the solution in the [pbl-updates](https://github.com/PeopleForBikes/brokenspoke-analyzer/tree/PBL-updates) branch:

This solution distinguishes true PBLs (which run adjacent and parallel to car traffic) from independent, off-street multi-use paths or trails. It accomplishes this by taking candidate cycleways, dropping sample points along their geometries, and checking if a sufficient fraction of those points are both strictly parallel to and within a tight distance threshold (10 meters) of an adjacent roadway.

lalver1:
# Pull-Request

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      change)
- [ ] Code cleanup / Refactoring
- [ ] Documentation

## Description

This PR updates the `bike_infra` script so that for each geometry explicitly tagged as `highway=cycleway`, and if it is completely within a 10m buffer of adjacent (UNIONed) car roads, treat it as a PBL (track). This identifies off-street paths that should actually be classified as PBLs instead. The buffer method is meant to replace calculating the angle between the potential PBL and the roadway, but still avoid misclassifying a bike path as a PBL when the path only crosses or "dead-ends" into a roadway. The value of 10 meters was found by testing a couple of cities and noticing that this distance is not too small or too large to cover how OSM mappers usually map PBLs as separate ways from the road.

### Examples

- Wauwatosa, WI (North 87th St)
  Current BNA  
  image

  This PR
  image

- Pasadena, CA (East Union St)
  Current BNA
  image

  This PR
  image

- Bend, OR (Olney Ave)
  Current BNA
  image
  
  This PR
  image

- Portland, OR (SW 4th Ave, east side of Naito Parkway)
  Current BNA  
  image

  
  This PR
  image

## Checklist

Fixes: #1194

### DIFF
--- a/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
+++ b/brokenspoke_analyzer/scripts/sql/features/bike_infra.sql
@@ -639,8 +639,9 @@ WHERE
 
 --------------------------------------------------------------------------------
 -- Protected bike lane (PBL) identification
--- Find geometries explicitly tagged as highway=cycleway, and if they are within
--- 10 meters of a road, treat them as PBLs (tracks). This identifies off-street
+-- For each geometry explicitly tagged as highway=cycleway, and if it is
+-- completely within a 10m buffer of adjacent (UNIONed) car roads,
+-- treat it as a PBL (track). This identifies off-street
 -- paths that should actually be classified as PBLs instead.
 --------------------------------------------------------------------------------
 UPDATE neighborhood_ways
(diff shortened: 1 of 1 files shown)


rgreinho:
Adding the `status/work-in-progress` to the PR so that the stale bot does not attempt to close it anymore.

PR #1215, opened 2026-09-03
Update international census block inclusion criteria

gstonecipher:
# Pull-Request

## Types of changes
This change updates the threshold census block inclusion from 50% to 25% for international cities.

The generated 1 km grid does not tend to align with city boundaries in the same way that US census data does. Given the larger size of the blocks, the 50% threshold means a lot of destinations that fall within the boundary are getting excluded, impacting the score.
Pasted Graphic 4

￼
By updating the threshold to 25%, a larger amount of destinations are captured without including census blocks that barely overlap. 

￼
Note: for testing, all previous input data for the city should be deleted, otherwise cached population data will be reused.

This example shows the City of London, Greater London, England.

uv run bna run-with compose england 'city of london' 'greater london' 

This change also addresses the issues from #1212.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      change)
- [ ] Code cleanup / Refactoring
- [ ] Documentation
- [ ] Infrastructure and automation

## Checklist

- [ ] I have updated the documentation accordingly
- [ ] I have updated the Changelog (if applicable)

Fixes: PeopleForBikes/PeopleForBikes.github.io#

PR #1216, opened 2026-09-03
Include WorldPop noData population blocks

gstonecipher:
# Pull-Request

## Types of changes
This change creates a continuous grid of census blocks including areas where WorldPop = NoData.

Previously, only WorldPop census blocks with population data were included. NoData blocks were filtered out. However, this also meant that destinations in those NoData areas were also ignored, because there was no way to count them within ‘connected_census_blocks’.

Pasted Graphic 2

This change creates a continuous grid of blocks, clips it to the boundary (with a buffer) to speed up processing, then vectorizes it pixel-by-pixel — since rasterio would otherwise merge adjacent blocks with equal values into one polygon. WorldPop data is then written into the new grid.

Pasted Graphic 3

This update may impact scores in cities with many destinations in NoData block areas. However, the BNA score for those blocks will have no impact due to population weighting.

This change only impacts international cities using WorldPop data.

Note: for testing, all previous input data for the city should be deleted, otherwise cached population data will be reused.

This shows the results of testing on Squamish, BC, Canada

uv run bna run-with compose canada 'squamish' 'british columbia'

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      change)
- [ ] Code cleanup / Refactoring
- [ ] Documentation

## Checklist

### DIFF
--- a/brokenspoke_analyzer/core/datasource.py
+++ b/brokenspoke_analyzer/core/datasource.py
@@ -263,11 +263,11 @@ class WorldPopAdapter(SourceAdapter):
         with rasterio.open(file_geotiff) as src:
             boundary_window = (
                 rasterio.windows.from_bounds(
-                minx, miny, maxx, maxy, transform=src.transform
+                    minx, miny, maxx, maxy, transform=src.transform
                 )
                 .round_offsets()
                 .round_shape()
-             )
+            )
             band_data = src.read(1, window=boundary_window)
             transform = src.window_transform(boundary_window)
 
--- a/brokenspoke_analyzer/core/datasource.py
+++ b/brokenspoke_analyzer/core/datasource.py
@@ -194,6 +194,7 @@ class WorldPopAdapter(SourceAdapter):
         self,
         country_iso_3166: str,
         year: int,
+        boundary_file: pathlib.Path,
         mirror: str | None = None,
     ) -> None:
         """
@@ -205,6 +206,7 @@ class WorldPopAdapter(SourceAdapter):
         super().__init__(mirror)
         self.country_iso_3166 = country_iso_3166
         self.year = year
+        self.boundary_file = boundary_file
 
     @staticmethod
     def key() -> str:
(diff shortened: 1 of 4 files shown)


rgreinho:
@gstonecipher : same question as in #1215, do we have to update something in the production cached data?

brokenspoke_analyzer/core/datasource.py
```diff
@@ -250,29 +251,63 @@ def prepare(self, datastore: pathlib.Path) -> None:
...
+            logger.warning(
+                f"Expected exactly 1 boundary geojson in {output_dir}, "
+                f"found {len(boundary_candidates)}; falling back to reading "
+                "the entire raster (this will be slow)."
+            )
+
         with rasterio.open(file_geotiff) as src:
-            # Read the population count as a numpy array
-            band_data = src.read(1)
+            if boundary_candidates and len(boundary_candidates) == 1:
```
rgreinho: `boundary_candidates` will never be `None` at this point so we can simplify the check:

```suggestion
            if len(boundary_candidates) == 1:
```

lalver1 (commented):
This looks good to me!

I'll propose a simplification to `boundary_candidates` later (since we already know the boundary file by the time `WorldPopAdapter.prepare` is called so there's probably no need for the logic in `boundary_candidates`, we can just pass in the boundary file in `prepare.py`'s `bna_store.download_worldpop`) but this works fine 👍

I rebased this branch onto `Update-international-block-overlap-ratio` (since it hasn't been merged yet but I wanted to see the effect it would have on this PR) and tested locally and saw very similar results for Squamish, BC, Canada:

Image

rgreinho:
Good catch here @lalver1!

> I'll propose a simplification to `boundary_candidates` later (since we already know the boundary file by the time `WorldPopAdapter.prepare` is called so there's probably no need for the logic in `boundary_candidates`, we can just pass in the boundary file in `prepare.py`'s `bna_store.download_worldpop`) but this works fine 👍

@gstonecipher There should never be a case where you have more than one boundary file. Did you face this issue? If so that may be another bug that we must address. Otherwise the boundary file is always `<data_dir>/<city>-<region>-<country>/<city>-<region>-<country>.geojson`

gstonecipher:
> @gstonecipher : same question as in #1215, do we have to update something in the production cached data?

I don't believe so. I haven't been clearing the local cache and it's been working fine - it's just the generated population.shp file in the 'data' folder that needs to be deleted, which I don't think is cached in the pipleline? At least, I don't see that file anywhere in the s3 bucket.

rgreinho:
> I don't believe so. I haven't been clearing the local cache and it's been working fine - it's just the generated population.shp file in the 'data' folder that needs to be deleted, which I don't think is cached in the pipleline? At least, I don't see that file anywhere in the s3 bucket.

Copy that! So we should be good to go in prod then 👍

gstonecipher:
> Good catch here @lalver1!
> 
> > I'll propose a simplification to `boundary_candidates` later (since we already know the boundary file by the time `WorldPopAdapter.prepare` is called so there's probably no need for the logic in `boundary_candidates`, we can just pass in the boundary file in `prepare.py`'s `bna_store.download_worldpop`) but this works fine 👍
> 
> @gstonecipher There should never be a case where you have more than one boundary file. Did you face this issue? If so that may be another bug that we must address. Otherwise the boundary file is always `<data_dir>/<city>-<region>-<country>/<city>-<region>-<country>.geojson`

Good call - I did not find this case, so this check can be deleted. This check is a relic of using Claude for assistance in coding and not getting rid of all the fluff - my bad!

rgreinho:
> Good catch here @lalver1!
> 
> > I'll propose a simplification to `boundary_candidates` later (since we already know the boundary file by the time `WorldPopAdapter.prepare` is called so there's probably no need for the logic in `boundary_candidates`, we can just pass in the boundary file in `prepare.py`'s `bna_store.download_worldpop`) but this works fine 👍
> 
> @gstonecipher There should never be a case where you have more than one boundary file. Did you face this issue? If so that may be another bug that we must address. Otherwise the boundary file is always `<data_dir>/<city>-<region>-<country>/<city>-<region>-<country>.geojson`

I think you should pass the boudary file as a new parameter to the datasource, ` boundary_file: pathlib.Path,`. then you could remove the checks and remove/simplify the fallback.

utils/cache-warmer.py
```diff
@@ -106,7 +106,11 @@ async def _run_downloads(
                 console.log(f"Downloading Worldpop data for {country.name}")
                 try:
                     await bna_store.download_worldpop(
-                        session, country.alpha_3, 2026, cache_only=cache_only
+                        session,
+                        country.alpha_3,
+                        2026,
+                        pathlib.Path(),
```
rgreinho, line 113: Nitpicky, but can you add a comment saying something like`# Unused: prepare() only runs when cache_only=False.` Just asking because I had to read it multiple times to makes sure it would not break anything, and I am sure in 2 weeks I will have forgotten why.

gstonecipher:
@rgreinho I think this implements the fix - because the cache-warmer uses the 'download_worldpop' function, I had to update that too. I'm not as familiar with the cache-warmer, so please let me know if this should be implemented differently

rgreinho (approved):
Looks good to me! 👍

